Issue #4, opened 2020-05-05
Implement common rules related to error types


HoucemKacem:
Based on architecture set up in done https://github.com/stellio-hub/stellio-context-broker/commit/bbf536e9cd60cce223d68b6dc16fab3d33f274c9

Some of the rules are already implemented, the task consists in implementing all the missing ones, all the code should go into the shared library.

PR #9, opened 2020-05-06
Feature/4 implement common rules related to error types

HoucemKacem:
- feat(shared): render appropriate error responses format #4 
- feat(entity): integrate common rules for handling errors #4
- feat(subscription): integrate common rules for handling errors #4
- feat: send the right format of error responses from handlers


shared/src/main/kotlin/com/egm/stellio/shared/model/ErrorResponse.kt
```diff
@@ -0,0 +1,19 @@
+package com.egm.stellio.shared.model
```
bobeal, line 1: A better approach is to have something like (it is of course to be completed and adjusted) :

```
sealed class ErrorResponse(
    val type: ErrorType,
    val title: String,
    val detail: String
)

data class BadRequestDataException(val detailedMessage: String): ErrorResponse(ErrorType.BAD_REQUEST_DATA, "", detailedMessage)
```

It has the advantages:
- To avoid redundancy and boilerplate code when raising exceptions
- You keep the control of the possible list of exceptions (list that is fixed by NGSI-LD specification)

shared/src/main/kotlin/com/egm/stellio/shared/util/ApiUtils.kt
```diff
@@ -40,6 +41,15 @@ object ApiUtils {
         mapper.readValue(content, Notification::class.java)
 }
 
+fun generatesProblemDetails(response: ErrorResponse): String {
```
bobeal, line 44: Is it needed ? Default JSON serialization of an ErrorResponse does not fit the needs ?

shared/src/main/kotlin/com/egm/stellio/shared/model/ErrorResponse.kt
bobeal: same principle may probably be applied to the exceptions hierarchy

bobeal:
Also, make a review of the uses of the `detail` field in error responses.

As per https://tools.ietf.org/html/rfc7807, it is optional ("The "detail" member, if present, ought to focus on helping the client correct the problem, rather than giving debugging information"), so, 
If filled, ensure it gives helpful information to the client, either remove it (for instance, "Entity not found" on a 404 does not bring much)

PR #17, opened 2020-05-13
Feature/16 setup testcontainers in entity service

HoucemKacem:


entity-service/build.gradle.kts
```diff
@@ -21,10 +21,7 @@ dependencies {
     developmentOnly("org.springframework.boot:spring-boot-devtools")
 
     testImplementation("org.hamcrest:hamcrest:2.1")
-
-    testRuntimeOnly("org.neo4j:neo4j-ogm-embedded-driver")
-    testRuntimeOnly("org.neo4j:neo4j-ogm-embedded-native-types")
-    testRuntimeOnly("org.neo4j:neo4j:3.5.12")
+    testImplementation("org.testcontainers:neo4j:1.14.1")
```
bobeal, line 24: Is it normal that the version does not match the one from testcontainers itself (1.12.3) ?

bobeal: And do we need this even if we are launching the containers via docker-compose?

HoucemKacem: actually we don't, i'll remove the dependency.

PR #44, opened 2020-05-22
Feature/27 integrate testcontainers in all modules

HoucemKacem:


entity-service/src/test/kotlin/com/egm/stellio/entity/config/TestContainersConfiguration.kt
```diff
@@ -0,0 +1,30 @@
...
+import org.springframework.boot.test.context.TestConfiguration
+import org.springframework.context.annotation.Bean
+
+@TestConfiguration
+class TestContainersConfiguration {
+
+    private val DB_USER = "neo4j"
+    private val DB_PASSWORD = "neo4j_password"
+
+    object EntityServiceTestContainers : TestContainers("neo4j", 7687) {
```
vraybaud, line 13: Shouldn't the name be something like Neo4JTestContainers ?

search-service/src/test/kotlin/com/egm/stellio/search/config/TestContainersConfiguration.kt
```diff
@@ -0,0 +1,45 @@
...
+            return instance.getServicePort(serviceName, servicePort)
+        }
+
+        fun getPostgresqlUri(): String {
+            return "jdbc:postgresql://" + getPostgresqlHost() + ":" + getPostgresqlPort() + '/'
+        }
+    }
+
+    @Bean
+    fun configuration(): ConnectionFactory {
```
vraybaud, line 33: The name of this method should be connectionFactory, right?

search-service/src/test/kotlin/com/egm/stellio/search/config/TestContainersConfiguration.kt
```diff
@@ -0,0 +1,45 @@
...
+import org.springframework.context.annotation.Bean
+
+@TestConfiguration
+class TestContainersConfiguration {
+
+    private val DB_NAME = "stellio_search"
+    private val DB_USER = "stellio_search"
+    private val DB_PASSWORD = "stellio_search_db_password"
+
+    object SearchServiceTestContainers : TestContainers("postgres", 5432) {
```
vraybaud, line 17: Maybe a PostgresConfiguration could be in shared with 3 methods : 
getDBName, getDBUser, getDBPassword to define 
```java
abstract class TestContainersConfiguration {

    abstract fun getDBName()
    abstract fun getDBUser()
    abstract fun getDBPassword()

    object PostgresTestContainers : TestContainers("postgres", 5432) {
Attach files by dragging & dropping, selecting or pasting them.

        fun getPostgresqlHost(): String {
            return instance.getServiceHost(serviceName, servicePort)
        }

        fun getPostgresqlPort(): Int {
            return instance.getServicePort(serviceName, servicePort)
        }

        fun getPostgresqlUri(): String {
            return "jdbc:postgresql://" + getPostgresqlHost() + ":" + getPostgresqlPort() + '/'
        }
    }

    @Bean
    fun configuration(): ConnectionFactory {
        val options = ConnectionFactoryOptions.builder()
            .option(ConnectionFactoryOptions.DATABASE, getDBName())
            .option(ConnectionFactoryOptions.HOST, SearchServiceTestContainers.getPostgresqlHost())
            .option(ConnectionFactoryOptions.PORT, SearchServiceTestContainers.getPostgresqlPort())
            .option(ConnectionFactoryOptions.USER, getDBUser())
            .option(ConnectionFactoryOptions.PASSWORD, getDBPassword())
            .option(ConnectionFactoryOptions.DRIVER, "postgresql")
            .build()

        return ConnectionFactories.get(options)
    }
}
```

entity-service/src/test/kotlin/com/egm/stellio/entity/config/TestContainersConfiguration.kt
HoucemKacem: Because it uses the docker-compose for entity-service but we are just exposing the neo4j service (what we need for now).
maybe, later, we will need to expose other services.
That's why i see it as EntityServiceTestContainers rather than Neo4JTestContainers

search-service/src/test/kotlin/com/egm/stellio/search/config/TestContainersConfiguration.kt
HoucemKacem: It's more significant yes.

search-service/src/test/kotlin/com/egm/stellio/search/config/TestContainersConfiguration.kt
HoucemKacem: Maybe this can avoid some code duplication but i think it's better to have separate configurations per service.
Open question

shared/build.gradle.kts
```diff
@@ -12,5 +12,7 @@ the<DependencyManagementExtension>().apply {
 }
 
 dependencies {
+    implementation("org.testcontainers:testcontainers:1.12.3")
```
bobeal, line 15: I understand why `implementation` and not `testImplementation` ... but I'm a bit worried about it, because the dependency (and its dependencies if any) will now be part of the produced jars.

bobeal: https://docs.gradle.org/current/userguide/java_testing.html#sec:java_test_fixtures looks promising

PR #46, opened 2020-05-25
Feature/45 add web flux tests in controllers testing

HoucemKacem:


entity-service/src/main/kotlin/com/egm/stellio/entity/config/WebSecurityConfig.kt
```diff
@@ -34,4 +39,9 @@ class WebSecurityConfig {
 
         return http.build()
     }
+
+    @Bean
+    fun jwtDecoder(): ReactiveJwtDecoder {
```
bobeal, line 44: Explain why you had to do this. And I'm not in favor of modifying something in the main source tree for testing purposes only.

HoucemKacem: Since WebFluxTest cannot detect the custom security configuration, we need to Import it explicitly.
The tests failed with (consider defining a bean of type ReactiveJwtDecoder in your configuration).
That's why i configured the bean.
We need to import the security configuration (in order to disable csrf) otherwise tests will fail with the status 403 forbidden.
We have another option: Define WebSecurityTestConfig in wich we disable csrf but we cannot test the anonymous access.

entity-service/src/main/kotlin/com/egm/stellio/entity/web/EntityOperationHandler.kt
```diff
@@ -28,7 +28,6 @@ class EntityOperationHandler(
      */
     @PostMapping("/create", consumes = [MediaType.APPLICATION_JSON_VALUE, JSON_LD_CONTENT_TYPE])
     fun create(@RequestBody body: Mono<String>): Mono<ResponseEntity<*>> {
-
```
bobeal, line 31: remove it from the commit (it is just noise)

PR #48, opened 2020-05-27
feat(entity): add support for entity attribute deletion #28

HoucemKacem:


entity-service/src/main/kotlin/com/egm/stellio/entity/repository/Neo4jRepository.kt
```diff
@@ -175,6 +175,48 @@ class Neo4jRepository(
         return Pair(queryStatistics.nodesDeleted, queryStatistics.relationshipsDeleted)
     }
 
+    @Transactional
+    fun deleteEntityProperty(entityId: String, propertyName: String): Pair<Int, Int> {
```
HoucemKacem, line 179: We already have deletePropertyFromEntity and deleteRelationshipFromEntity functions that are used in the entityService.
But both does not delete the children (properties and relationships) of the attribute.

entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityService.kt
```diff
@@ -602,6 +602,16 @@ class EntityService(
         return neo4jRepository.deleteEntity(entityId)
     }
 
+    fun deleteEntityAttribute(entityId: String, attributeName: String, contextLink: String): Pair<Int, Int> {
```
vraybaud, line 605: Is return type needed ?

entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityService.kt
```diff
@@ -602,6 +602,16 @@ class EntityService(
         return neo4jRepository.deleteEntity(entityId)
     }
 
+    fun deleteEntityAttribute(entityId: String, attributeName: String, contextLink: String): Pair<Int, Int> {
+        val expandedAttributeName = expandJsonLdKey(attributeName, contextLink)!!
+
+        if (neo4jRepository.hasPropertyOfName(entityId, expandedAttributeName))
```
vraybaud, line 608: I think it could be straightforward if you just do a method getAttributeOfSubject (see getPropertyOfSubject) and work with Attribute type and then delete all linked properties with deleteAll(list)

What do you think ?

entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityService.kt
HoucemKacem: No, it should be removed.

entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityService.kt
HoucemKacem: I'm not sure, giving an entityId and an attribute name, we can have a generic method that returns the attribute instance.
Moreover, i don't see how we can delete this attribute and its attributes with the deleteAll() function.

entity-service/src/main/kotlin/com/egm/stellio/entity/repository/Neo4jRepository.kt
bobeal: Should this new `deleteEntityProperty` replace the existing `deletePropertyFromEntity` where it is used (and same for relationships)?

entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityService.kt
vraybaud: I don't know if that's possible I will check it out but I got the idea from the getPropertyOfSubject already written.

entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityService.kt
bobeal: The current return type is not very meaningful, but in a FP way of thinking, it is recommended to return something (that you can do something with of course). A better return type would be something like Either<Boolean, Exception> (to be improved later).

entity-service/src/main/kotlin/com/egm/stellio/entity/repository/Neo4jRepository.kt
HoucemKacem: deletePropertyFromEntity and deleteRelationshipFromEntity are used in the appendEntityAttributes service function.
More precisely when we are appending an existing attribute with (allowing override option), so we are removing the old one and attaching the new one to the entity.
I think we should replace them with the new deleteEntityProperty and deleteEntityRelationship, since we don't need the children attributes anymore if the attribute itself was deleted.

bobeal: Go for it then

entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityService.kt
```diff
@@ -602,6 +602,20 @@ class EntityService(
         return neo4jRepository.deleteEntity(entityId)
     }
 
+    fun deleteEntityAttribute(entityId: String, attributeName: String, contextLink: String): Boolean {
+        val expandedAttributeName = expandJsonLdKey(attributeName, contextLink)!!
+
+        if (neo4jRepository.hasPropertyOfName(entityId, expandedAttributeName)) {
+            neo4jRepository.deleteEntityProperty(entityId, expandedAttributeName)
+            return true
```
bobeal, line 610: you could return something like `deleteEntityProperty(...) >= 1` (and then use the boolean result in the handler to decide on the response : `true` -> 204, `false` -> 400 or 500)

bobeal: Currently, whatever the result of the delete operation, we say it's successful (generally, errors will probably result in an exception but one never knows ...)

Issue #21, opened 2020-05-14
Add entityOperations/upsert endpoint


vraybaud:
As per clause 6.15.3.1 of the specification

PR #55, opened 2020-06-04
feat(entity): support entityOperations/upsert replace option #21

vraybaud:
end of entityOperations/upsert feature

### DIFF
--- a/entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityOperationService.kt
+++ b/entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityOperationService.kt
@@ -14,6 +14,7 @@ import org.jgrapht.Graphs
 import org.jgrapht.graph.DefaultEdge
 import org.jgrapht.graph.DirectedPseudograph
 import org.springframework.stereotype.Component
+import org.springframework.transaction.annotation.Transactional
 import kotlin.streams.toList
 
 /**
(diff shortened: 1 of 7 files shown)


entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityService.kt
```diff
@@ -613,7 +613,8 @@ class EntityService(
     }
 
     fun updateEntityLastMeasure(observation: Observation) {
-        val observingEntity = neo4jRepository.getObservingSensorEntity(observation.observedBy, EGM_VENDOR_ID, observation.attributeName)
```
bobeal, line 616: If possible, remove this from the commit (that's the only diff in the file and it's only formatting)

entity-service/src/test/kotlin/com/egm/stellio/entity/web/EntityOperationHandlerTests.kt
```diff
@@ -282,6 +282,47 @@ class EntityOperationHandlerTests {
...
+        )
+
+        webClient.post()
+            .uri("/ngsi-ld/v1/entityOperations/upsert")
+            .header("Link", "<$aquacContext>; rel=http://www.w3.org/ns/json-ld#context; type=application/ld+json")
+            .bodyValue(jsonLdFile)
+            .exchange()
+            .expectStatus().isOk
+            .expectBody().json(
+                "{\n" +
```
bobeal, line 315: multiline string ? :)

entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityOperationService.kt
```diff
@@ -103,27 +124,42 @@ class EntityOperationService(
...
     }
 
+    /*
+     * Transactional because it should not delete entity attributes if new ones could not be appended.
+     */
+    @Transactional(rollbackFor = [BadRequestDataException::class])
+    @Throws(BadRequestDataException::class)
+    private fun replaceEntity(entity: ExpandedEntity): Either<BatchEntityError, String> {
+        neo4jRepository.deleteEntityAttributes(entity.id)
+        entityService.appendEntityAttributes(entity.id, entity.attributes, false)
```
bobeal, line 140: I'm wondering if a check that updated.size == entity.attributes.size would make sense?

vraybaud: Indeed it would make sense, and I think it should be done.

Issue #37, opened 2020-05-15
Upgrade to Neo4j 4.0


bobeal:
Process is fairly easy: https://neo4j.com/blog/graphcast-migrating-from-neo4j-3-5-to-4-0

Basically:
- Update binaries
- Move DB data and rename default database
- Activate migration in neo4j conf
- Ensure JDK11+ (we already do)

bobeal:
And the deployment procedure : https://neo4j.com/docs/operations-manual/current/upgrade/deployment-upgrading/

bobeal:
Also be sure we are using compatible versions for neo4j-ogm, neo4j java driver, ... : https://github.com/neo4j/neo4j-ogm/wiki/Versions#recommended-versions

PR #58, opened 2020-06-05
feat(entity): upgrade to Neo4j 4.0 + update cypher queries syntax #37

HoucemKacem:
Concerning versions compatibility with Neo4j 4.0, project dependencies (Neo4j-OGM and Spring data Neo4j) are in the same version as for Neo4j 3.5


entity-service/src/main/kotlin/com/egm/stellio/entity/repository/EntityRepository.kt
```diff
@@ -8,36 +8,32 @@ import org.springframework.stereotype.Repository
 @Repository
 interface EntityRepository : Neo4jRepository<Entity, String> {
 
-    @Query("""
-        MATCH (entity:Entity { id: {id} })
-        RETURN entity
-    """)
+    @Query("MATCH (entity:Entity { id: \$id })" +
```
vraybaud, line 11: You cannot do multi-line strings anymore?

HoucemKacem: From the documentation:
A raw string is delimited by a triple quote ("""), contains no escaping and can contain newlines and any other character.
So we need to use the standard string.

Or, there is another option (use the literal representation), for this example:
"""MATCH (entity:Entity { id: ${'$'}id }) RETURN entity"""

HoucemKacem:
Steps to upgrade to Neo4j 4.x (with docker):
- Backup the mounted volume 
- Update image version
- Set NEO4J_dbms_allow__upgrade environment variable to true (This enable upgrading the store format to the 4.x structure)
- Update the database name in the mounted volume from graph.db to neo4j (Since the 4.x version add support for multiple databases that follow a naming convention without special characters including dots)   
- Recreate the Neo4j container (docker-compose up) 
When the upgrade has finished, dbms.allow_upgrade should be set to false or be removed.
In case of connection problem, please consider setting the NEO4J_dbms_recovery_fail__on__missing__files enviroment variable to false, removing the neostore.transaction.db.0 file under the new neo4j folder and recreating the container.

Steps to upgrade to Neo4j 4.x (without docker):
- Download Neo4j 4.x 
- Copy data folder content from the old Neo4j directory to the new Neo4j 4.x data directory
- Edit config file under conf/neo4j.conf, set NEO4J_dbms_allow__upgrade to true  (This enable upgrading the store format to the 4.x structure)
- Rename graph.db database folder to neo4j (Since the 4.x version add support for multiple databases that follow a naming convention without special characters including dots)   
- Start Neo4j 4.x

For browser connection to Neo4j, please make sure that the Connect URL uses the bolt protocol.

bobeal:
> Steps to upgrade to Neo4j 4.x (with docker):

It misses the commands to run and / or exact things to modify for each instruction. Think of someone that installed and uses Stellio, but does not know (and does not want to) the inner technical details of Stellio and neo4j. 

>     * Update the database name in the mounted volume from graph.db to neo4j (Since the 4.x version add support for multiple databases that follow a naming convention without special characters including dots)

Can't we name it stellio (or stellio-entity) instead?

> Steps to upgrade to Neo4j 4.x (without docker):

We always use neo4j in Docker, so we don't need to add instructions for neo4j not running in Docker.

HoucemKacem:
Yes, we can name it stellio but stellio-entity is not supported since special characters are not supported.
 
Steps to upgrade to Neo4j 4.x:

- Backup the mounted volume
- Update image version by choosing neo4j:4.0 as image name 
- In docker-compose environment section of the neo4j service, add NEO4J_dbms_allow__upgrade=true (This enable upgrading the store format to the 4.x structure) and add NEO4J_dbms_default__database=stellio
- Update the database name in the mounted volume from graph.db to stellio (Since the 4.x version add support for multiple databases that follow a naming convention without special characters including dots), the update can be done by renaming the graph.db folder under data/databases folder  
- Recreate the Neo4j container by running docker-compose up
When the upgrade has finished, dbms.allow_upgrade should be set to false or be removed.
In case of connection problem, please consider setting the NEO4J_dbms_recovery_fail__on__missing__files enviroment variable to false, removing the neostore.transaction.db.0 file under the new neo4j folder and recreating the container by running docker-compose up.

For browser connection to Neo4j, please make sure that the Connect URL uses the bolt protocol.

HoucemKacem:
This note describes the necessary steps to migrate a single instance from Neo4j version 3.5 to Neo4j version 4.0.x

- Backup:
The Data generated and used by the neo4j container to run the neo4j database instance is persisted with a docker volume (declared as stellio-entity-neo4j-storage in the entity-service docker-compose.yml).         
We need to make a backup of this volume that can be used if the migration goes wrong.
To list the docker volumes: docker volume ls
We will use the entity-service_stellio-entity-neo4j-storage volume (please consider working with your own volume name if you changed it in the entity-service docker-compose.yml). 
To locate the volume mountpoint: docker volume inspect entity-service_stellio-entity-neo4j-storage 
The mountpoint is the path to the _data directory generated by the neo4j container.
Make a copy of this directory. 


- Migration:
We will update the neo4j service in the docker-compose.yml of the entity-service . 
1- Update the service image: Replace neo4j:3.5.14 by neo4j:4.0
2- Update the service environment: 
	- add NEO4J_dbms_allow__upgrade=true (This enable upgrading the store format to the 4.x structure).
	- add NEO4J_dbms_default__database=stellio (This set stellio database as a default database). 
  
3- Under the _data mounted folder, update the database name by renaming the graph.db folder under databases folder to stellio (Please consider that the 4.x version add support for multiple databases that follow a naming convention without special characters including dots).
4- In the entity service, recreate the neo4j container by running docker-compose up, you can check in the logs that the migration was done successfully. 
5- Once finished, dbms.allow_upgrade set previously in the docker-compose file should be set to false or be removed.

In case of connection problem, please consider setting the NEO4J_dbms_recovery_fail__on__missing__files enviroment variable to false, removing the neostore.transaction.db.0 file under the new stellio folder and recreating the container by running docker-compose up.

For browser connection to Neo4j, please make sure that the Connect URL uses the bolt protocol.

bobeal:
For the migration notes, after checking the practices in use:
- Add a page in the [Stellio site](https://github.com/stellio-hub/stellio-docs) (cf what is done in https://fiware-orion.readthedocs.io/en/2.4.0/admin/upgrading_crossing_2-2-0/index.html)
- Reference the page in the Draft release notes for the next version (cf what is done in https://github.com/telefonicaid/fiware-orion/releases/tag/2.2.0)

bobeal:
> This note describes the necessary steps to migrate a single instance from Neo4j version 3.5 to Neo4j version 4.0.x

When migrating to the site, pay attention to the formatting (code snippets, commands, ...), it greatly improves the readability.

>       The Data generated and used by the neo4j container to run the neo4j database instance 

Just "the neo4j database is persisted in ..." should be enough :)

>       3- Under the _data mounted folder, update the database name by renaming the graph.db folder under databases folder to stellio (stellio is the default database name set with the NEO4J_DEFAULT_DATABASE variable in entity service .env file).

You could illustrate with a sample command (`cp -r $neo4j_volume/... $neo4j_volume/...`) to make it clearer.

HoucemKacem:
For the release note: 

**Important note: From version 0.7.0 and later, a migration procedure on existing data has to be applied see details [here](https://stellio.readthedocs.io/en/latest/source/migrations/neo4j-migrations.html)**

bobeal:
I'm a little bit surprised (and worried) that we are not telling to entity-service to use the stellio database in neo4j (it is only configured in neo4j as the default one).

HoucemKacem:
> I'm a little bit surprised (and worried) that we are not telling to entity-service to use the stellio database in neo4j (it is only configured in neo4j as the default one).

I checked that the entity service is connected to the stellio database.
Maybe because it's the default database.

bobeal:
> > I'm a little bit surprised (and worried) that we are not telling to entity-service to use the stellio database in neo4j (it is only configured in neo4j as the default one).
> 
> I checked that the entity service is connected to the stellio database.
> Maybe because it's the default database.

Sure but it is dangerous, isn't it? Isn't there a way to specify the database to use at entity service level?

HoucemKacem:
> Sure but it is dangerous, isn't it? Isn't there a way to specify the database to use at entity service level?

Yes, we can specify the database to use at the service level #58 63780b4410c1d414420212c5e7d12d89fdffa434

entity-service/src/main/kotlin/com/egm/stellio/entity/config/Neo4jConfiguration.kt
```diff
@@ -1,21 +1,24 @@
...
 @Configuration
 @EnableNeo4jRepositories(basePackages = ["com.egm.stellio.entity.repository"])
-@EnableTransactionManagement
-class Neo4jConfiguration(
-    private val sessionFactory: SessionFactory
-) {
+class Neo4jConfiguration {
 
     @Bean
-    fun transactionManager(): Neo4jTransactionManager {
```
bobeal, line 18: what is the rationale behind removing the transaction manager (and management)?

HoucemKacem: similar to [this](https://github.com/neo4j/neo4j-java-driver-spring-boot-starter/issues/10) issue, we just need to configure the OGM starter.
Otherwise, it would be like a double configuration and there is a cyclic dependency between the manual OGM configuration and the sessionFactory.
I tried to keep the Neo4jTransactionManager but the application failed to start due to the mentionned dependencies conflict.

bobeal: well, not exactly the same, because we were not creating a new session factory (but using the existing one to create the transaction manager, based on what is said in [SDN manual](https://docs.spring.io/spring-data/data-neo4j/docs/current/reference/html/#reference.getting_started.spring-configuration): "Note that you will have to activate @EnableTransactionManagement explicitly to get annotation based configuration at facades working as well as define an instance of this Neo4jTransactionManager with the bean name transactionManager".

but if you can guarantee that transactions work as expected at the service level, I'll buy it.

entity-service/src/main/kotlin/com/egm/stellio/entity/config/Neo4jConfiguration.kt
```diff
@@ -1,21 +1,38 @@
...
+        return SessionFactory(ogmConfiguration(properties), "com.egm.stellio.entity.model")
+    }
+
+    @Bean
+    fun ogmConfiguration(properties: Neo4jProperties): org.neo4j.ogm.config.Configuration {
+        return org.neo4j.ogm.config.Configuration.Builder()
+            .uri(properties.uri)
+            .credentials(properties.username, properties.password)
+            .useNativeTypes()
+            .database("stellio")
```
vraybaud, line 29: shouldn't this be configurable, or the same as default database ?

HoucemKacem: Stellio is the default database

Issue #56, opened 2020-06-04
Format all files aligned with ktlint and share the format settings


vraybaud:
All files are not formatted the same way.
Currently we don't have any format configuration shared for the developers to matter the minimum about formatting. 

We should format all files and share an Intellij configuration for developers to contribute better on the project.

PR #59, opened 2020-06-08
refactoring: format all code #56

vraybaud:
- Import orders are now compliant with the Ktlint Intellij Settings. If you do ```ktlint```, you'll find out "Imports must be ordered in lexicographic order without any empty lines in-between" errors, there are a known issue of ktlint <= 0.36.0 : https://github.com/pinterest/ktlint/issues/527 which is fixed in v0.37.0, which requires an update from org.jlleitschuh.gradle.ktlint plugin.

- Indentations are ajusted, removing useless indentations that were in many files.
- Hard wrapping at 120 characters to ease the code review.
- Add new line at end of file (see POSIX definition of a line)

Please if you see something strange, do comment it. (I could have missed something)


subscription-service/src/test/kotlin/com/egm/stellio/subscription/service/NotificationServiceTests.kt
```diff
@@ -212,15 +257,17 @@ class NotificationServiceTests {
...
-            .willReturn(ok()))
+        stubFor(
+            post(urlMatching("/notification"))
+                .willReturn(ok())
+        )
 
         StepVerifier.create(notificationService.callSubscriber(subscription, listOf(parsedEntity)))
             .expectNextMatches {
                 it.first.id == subscription.id &&
-                it.second.subscriptionId == subscription.id &&
```
HoucemKacem, line 221: indentation seems to be not good here (and below).

HoucemKacem:
Do we need to configure something in Intellij ?
From now on how can we check the code style ?

subscription-service/src/test/kotlin/com/egm/stellio/subscription/service/NotificationServiceTests.kt
vraybaud: This is to clearly identify the continuation of the expression, so you clearly see at first sight that this is all a single expression.

Maybe in this case it seems useless, but you could have cases like that : 
```kotlin
val a = it.size == 1 &&
it[0].second.subscriptionId == subscription.id &&
it[0].second.data.size == 1 &&
it[0].third
a
```
There it becomes more relevant to have this indent to more easily detect the what is variable a and what is returned.

vraybaud:
Indeed I'll share a common settings file to import for Intellij IDEA users to properly make changes without braking the format of every files. (tbd where in the project)

entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityService.kt
```diff
@@ -164,29 +193,30 @@ class EntityService(
         values.filterValues {
             it[0] is Map<*, *>
         }
-        .mapValues {
-            expandValueAsMap(it.value)
-        }
-        .filter {
-            !(NGSILD_PROPERTIES_CORE_MEMBERS.plus(NGSILD_RELATIONSHIPS_CORE_MEMBERS)).contains(it.key) &&
+            .mapValues {
```
bobeal, line 196: I find weird this extra indent. I guest it is considered like a continuation indent but ...

entity-service/src/test/kotlin/com/egm/stellio/entity/web/EntityHandlerTests.kt
```diff
@@ -274,7 +312,8 @@ class EntityHandlerTests {
             .header(HttpHeaders.CONTENT_TYPE, MediaType.APPLICATION_JSON_VALUE)
             .exchange()
             .expectStatus().isOk
-            .expectBody().json("{\"testedAt\":{\"type\":\"Property\",\"value\":{\"type\":\"Time\",\"@value\":\"11:20:30\"}},\"@context\":\"$NGSILD_CORE_CONTEXT\"}")
+            .expectBody()
+            .json("{\"testedAt\":{\"type\":\"Property\",\"value\":{\"type\":\"Time\",\"@value\":\"11:20:30\"}},\"@context\":\"$NGSILD_CORE_CONTEXT\"}")
```
bobeal, line 316: This one bothers me because we lose the fact that `.json(...)` applies to the body (so is directly related to the previous line). Strangely enough, it is correctly formatted in the next diff.

entity-service/src/test/kotlin/com/egm/stellio/entity/web/EntityHandlerTests.kt
```diff
@@ -376,17 +430,25 @@ class EntityHandlerTests {
         val entityId = "urn:ngsi-ld:BreedingService:0214"
 
         every { entityService.exists(any()) } returns true
-        every { entityService.appendEntityAttributes(any(), any(), any()) } throws BadRequestDataException("@type not found")
+        every {
+            entityService.appendEntityAttributes(
+                any(),
```
bobeal, line 435: The hard wrap is for any method call?

search-service/src/test/kotlin/com/egm/stellio/search/service/AttributeInstanceServiceTests.kt
```diff
@@ -84,8 +94,10 @@ class AttributeInstanceServiceTests : TimescaleBasedTests() {
 
         (1..10).forEach { _ -> attributeInstanceService.create(gimmeAttributeInstance()).block() }
 
-        val temporalQuery = TemporalQuery(emptyList(), TemporalQuery.Timerel.AFTER, Instant.now().atZone(ZoneOffset.UTC).minusHours(1),
-            null, null, null)
+        val temporalQuery = TemporalQuery(
```
bobeal, line 97: Strange, it is not consistent with formatting at https://github.com/stellio-hub/stellio-context-broker/pull/59/files#diff-8adafc7784840c932f2aeead03149087R69

shared/src/main/kotlin/com/egm/stellio/shared/util/NgsiLdParsingUtils.kt
```diff
@@ -220,7 +232,7 @@ object NgsiLdParsingUtils {
     }
 
     fun extractShortTypeFromPayload(payload: Map<String, Any>): String =
-        // TODO is it always after a '/' ? can't it be after a '#' ? (https://redmine.eglobalmark.com/issues/852)
+    // TODO is it always after a '/' ? can't it be after a '#' ? (https://redmine.eglobalmark.com/issues/852)
```
bobeal, line 235: Hmm, why isn't it aligned with the comment below (and the line after)?

search-service/src/main/kotlin/com/egm/stellio/search/web/TemporalEntityHandler.kt
```diff
@@ -38,8 +52,12 @@ class TemporalEntityHandler(
...
      */
     @PostMapping("/{entityId}/attrs", consumes = [MediaType.APPLICATION_JSON_VALUE, JSON_LD_CONTENT_TYPE])
-    fun addAttrs(@RequestHeader httpHeaders: HttpHeaders, @PathVariable entityId: String, @RequestBody body: Mono<String>):
-            Mono<ResponseEntity<*>> {
+    fun addAttrs(
+        @RequestHeader httpHeaders: HttpHeaders,
+        @PathVariable entityId: String,
+        @RequestBody body: Mono<String>
+    ):
+        Mono<ResponseEntity<*>> {
```
bobeal, line 60: It should be on the previous line (after the `:`), at least it is in others

entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityService.kt
bobeal: I would be more comfortable with something like:

```
values
    .map {
    }
    .filter {
    }
```

bobeal: Rather than

```
values.map {
}
    .filter {
    }
```

bobeal:
> Indeed I'll share a common settings file to import for Intellij IDEA users to properly make changes without braking the format of every files. (tbd where in the project)

Add this in the PR (with instructions in the README)

search-service/src/test/kotlin/com/egm/stellio/search/service/AttributeInstanceServiceTests.kt
vraybaud: Yes indeed this is an error, I guess the forced newline made the compiler fail

entity-service/src/test/kotlin/com/egm/stellio/entity/web/EntityHandlerTests.kt
vraybaud: You are right, I agree that in this kind of cases you need to force the formatting. The fact that in the next diff it was correctly formatted is because there were new line before for the first arg, there not.

entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityService.kt
vraybaud: I don't think continuation indent applies to braces... 

I think expected result is : 
```kotlin
values.filterValues {
    it[0] is Map<*, *>
}.mapValues {
    expandValueAsMap(it.value)
}.filter {
    !(NGSILD_PROPERTIES_CORE_MEMBERS.plus(NGSILD_RELATIONSHIPS_CORE_MEMBERS)).contains(it.key) &&
        isAttributeOfType(it.value, NGSILD_PROPERTY_TYPE)
}.forEach { entry ->
    if (!neo4jRepository.hasPropertyOfName(subject.id, entry.key))
        throw BadRequestDataException("Property ${entry.key.extractShortTypeFromExpanded()} does not exist")
    updatePropertyValues(subject.id, entry.key, entry.value)
}
```
But there were a newline placed in first line..
What do you think ?

entity-service/src/test/kotlin/com/egm/stellio/entity/web/EntityHandlerTests.kt
vraybaud: It is not. This one is a consequence of the hard wrap of the whole line : 
```kotlin
every { entityService.appendEntityAttributes(any(), any(), any()) } throws BadRequestDataException("@type not found")
``` 
Could be : 
```kotlin
every { entityService.appendEntityAttributes(any(), any(), any()) }
    throws BadRequestDataException("@type not found")
```
or : 
```kotlin
every { 
    entityService.appendEntityAttributes(any(), any(), any()) 
} throws BadRequestDataException("@type not found")
```
But Intellij seems to apply formatting rules for long lines to every elements in the line.

shared/src/main/kotlin/com/egm/stellio/shared/util/NgsiLdParsingUtils.kt
vraybaud: In my opinion this is caused by the function without braces, with braces it is correctly formatted.
Why is this happening ? I don't know. I couldn't find a way to arrange that.
Without the comments, it formats like that : 
```kotlin
fun extractShortTypeFromPayload(payload: Map<String, Any>): String =
        (payload["@type"] as List<String>)[0].substringAfterLast("/").substringAfterLast("#")
```
and with a multi line comment like that : 
```kotlin
fun extractShortTypeFromPayload(payload: Map<String, Any>): String =
    /* 
     * TODO do a clean implementation using info from @context
     * TODO is it always after a '/' ? can't it be after a '#' ? (https://redmine.eglobalmark.com/issues/852)
     */
    (payload["@type"] as List<String>)[0].substringAfterLast("/").substringAfterLast("#")
```
I think a common rule is to use multi line comment instead of having 2 separate comments. Should I replace it by a multi-line comment ?

vraybaud:
> > Indeed I'll share a common settings file to import for Intellij IDEA users to properly make changes without braking the format of every files. (tbd where in the project)
> 
> Add this in the PR (with instructions in the README)

I will add instructions in the README for this PR but we should do an issue (not priority) to create a CONTRIBUTING.md

entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityService.kt
bobeal: I'm fine with your expected result!

shared/src/main/kotlin/com/egm/stellio/shared/util/NgsiLdParsingUtils.kt
bobeal: OK

bobeal:
> > > Indeed I'll share a common settings file to import for Intellij IDEA users to properly make changes without braking the format of every files. (tbd where in the project)
> > 
> > 
> > Add this in the PR (with instructions in the README)
> 
> I will add instructions in the README for this PR but we should do an issue (not priority) to create a CONTRIBUTING.md

Yep, create the issue.

entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityService.kt
vraybaud: Fixed in d6d00ee57a34597b8636634c9f94a7731810dcac

PR #67, opened 2020-06-11
Fix/64 fix temporal values injection

HoucemKacem:


search-service/src/main/kotlin/com/egm/stellio/search/service/TemporalEntityAttributeService.kt
```diff
@@ -226,19 +226,19 @@ class TemporalEntityAttributeService(
...
-                // For this reason, a cast to Instant is needed (with UTC as ZoneOffset) to return the expected dateTime with Z format.
+                // The value is retrieved as offsetDateTime and converted to the current timezone using the system variable timezone.
+                // For this reason, a cast to Instant with UTC as ZoneOffset is needed to create a ZonedDateTime.
                 val valuesMap =
                     it.map {
                         if (it["value"] is Double)
                             TemporalValue(
                                 it["value"] as Double,
-                                (it["observed_at"] as ZonedDateTime).toInstant().atZone(ZoneOffset.UTC).toString()
+                                ZonedDateTime.parse(it["observed_at"].toString()).toInstant().atZone(ZoneOffset.UTC).toString()
```
bobeal, line 236: I still think this is quite cumbersome but if there is no other way ...

HoucemKacem: It can be done this way, OffsetDateTime.parse(it["observed_at"].toString()).atZoneSameInstant(ZoneOffset.UTC)

But the returned value is not in a textual representation: "yyyy-MM-dd'T'HH:mm:ss.SSSZ"

I tried to use the serializeObject function in ApiUtils (since it uses a mapper with SerializationFeature.WRITE_DATES_AS_TIMESTAMPS disabled), i got something like "\"yyyy-MM-dd'T'HH:mm:ss.SSSZ\""

bobeal: ok, let's stay on this for now

PR #69, opened 2020-06-11
feat(WIP): improve entity creation process #70

bobeal:


entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityService.kt
```diff
@@ -130,7 +114,7 @@ class EntityService(
         entity: Entity,
         propertyKey: String,
         propertyValues: Map<String, List<Any>>
-    ): Property {
+    ): String {
         val propertyValue = getPropertyValueFromMap(propertyValues, NGSILD_PROPERTY_VALUE)
```
vraybaud, line 118: Not related to improvement but readibility, maybe we could move the building part of the Property/Relationship in a builder ?

vraybaud: Or even at least in constructors or an Util class

entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityService.kt
```diff
@@ -130,7 +114,7 @@ class EntityService(
         entity: Entity,
         propertyKey: String,
         propertyValues: Map<String, List<Any>>
```
vraybaud, line 116: Could we have an ExpandedProperty or ExpandedAttribute to have the same utils method as for ExpandedEntity and get relationships/properties directly ?

entity-service/src/main/kotlin/com/egm/stellio/entity/repository/Neo4jRepository.kt
```diff
@@ -24,30 +23,38 @@ class Neo4jRepository(
...
+        val query = """
+            MATCH (subject { id: '$subjectId' }), (target:Entity { id: '$targetId' })
+            CREATE (subject)-[:HAS_OBJECT]->(r:Attribute:Relationship:`${relationship.type[0]}` ${'$'}props)-[:HAS_OBJECT]->(target)
+            CREATE (r)-[:$relationshipType]->(target)
+            RETURN r.id as id
+        """
+
+        return session.query(query, mapOf("props" to relationship.nodeProperties()))
+            .toList()
+            .map { it["id"] as String }
```
vraybaud, line 51: Better to get first() then to get ["id"] as String or i'm wrong?

bobeal: Yes indeed

entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityService.kt
bobeal: Yes, having this kind of abstraction for properties and relationships is planned for a near future

entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityService.kt
bobeal: Also

PR #72, opened 2020-06-16
feat: set default conf for actuator endpoints

bobeal:
- only activate health endpoint by default
- allow unauthenticated access to enabled endpoints (in production, use a security layer above)
- add exporter for Prometheus metrics

### DIFF
--- a/entity-service/src/main/resources/application.properties
+++ b/entity-service/src/main/resources/application.properties
@@ -11,8 +11,8 @@ spring.security.oauth2.resourceserver.jwt.jwk-set-uri = https://data-hub.eglobal
 
 server.port = 8082
 
-management.endpoints.web.exposure.include = health,info,metrics,bindings
-management.health.db.enabled = false
+management.endpoints.enabled-by-default = false
+management.endpoint.health.enabled = true
 
 application.authentication.enabled = false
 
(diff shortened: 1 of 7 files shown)


vraybaud:
What is this for ? Have micrometer metrics in Prometheus ?

bobeal:
Yes (cf commit message)

Le 16 juin 2020 à 13:38, Vincent <notifications@github.com> a écrit :

﻿

What is this for ? Have micrometer metrics in Prometheus ?

bobeal:
> What is this for ? Have micrometer metrics in Prometheus ?

Yes (cf commit message)

entity-service/src/main/resources/application.properties
```diff
@@ -11,8 +11,8 @@ spring.security.oauth2.resourceserver.jwt.jwk-set-uri = https://data-hub.eglobal
 
 server.port = 8082
 
-management.endpoints.web.exposure.include = health,info,metrics,bindings
-management.health.db.enabled = false
```
HoucemKacem, line 15: Does the added properties disable the db health check ?
Since we had a deployment problem related to this

bobeal: We had a problem with the migration tool you removed since ...

Issue #65, opened 2020-06-10
Fix error response when creating an entity having non-existent relationship object


HoucemKacem:
When creating an entity having a non-existent relationship object, the expected error response is a 400 with the detail (Target entity $objectId does not exist, create it first).

However, the error response is a 500 with the following payload
```
{
    "detail": "There has been an error during the operation execution",
    "title": "There has been an error during the operation execution",
    "type": "https://uri.etsi.org/ngsi-ld/errors/InternalError"
} 
```

Issue #41, opened 2020-05-20
Internal error when BadRequest exception occurs


vraybaud:
When doing that request :

POST /ngsi-ld/v1/entities HTTP/1.1
```json
{
  "id": "urn:ngsi-ld:Vehicle:A12388",
  "type": "Vehicle",
  "speed": {
    "type": "Property"
  },
  "@context": [
    "https://raw.githubusercontent.com/easy-global-market/ngsild-api-data-models/master/shared-jsonld-contexts/egm.jsonld",
    "https://raw.githubusercontent.com/easy-global-market/ngsild-api-data-models/master/aquac/jsonld-contexts/aquac.jsonld",
    "http://uri.etsi.org/ngsi-ld/v1/ngsi-ld-core-context.jsonld"
  ]
}
```

The response should be an error BadRequestData (400 http) like that, helping the user figure out what's wrong in the request :
```json
{
    "detail": "Key $NGSILD _PROPERTY_VALUE not found in $propertyValues").
    "type": "https://uri.etsi.org/ngsi-ld/errors/BadRequestData",
    "title": "The request includes input data which does not meet the requirements of the operation"
}
```

But today we get this : (500 http)
```json
{
    "detail": "There has been an error during the operation execution",
    "type": "https://uri.etsi.org/ngsi-ld/errors/InternalError",
    "title": "There has been an error during the operation execution"
}
```

**Investigation :**
This happens because the error occurs in a function with @Transactional annotation which seems to not pass correctly the exception.

bobeal:
Indeed, current code in ``NgsiLdParsingUtils` is not totally bulletproof (in other words, it expects a value and crashes miserably if none is found).

A refactoring has been started to improve validation and type safety (cf for instance `ExpandedEntity` in shared lib). 

We'll talk about it soon for the next steps.

vraybaud:
Found a related issue : https://github.com/spring-projects/spring-framework/issues/22412

vraybaud:
Also another bug somehow close to this issue : 
@Transactional() alone in kt only rollbacks the transaction for unchecked Exceptions
It should be modified to 
    @Transactional(rollbackFor = [BadRequestDataException::class])
for example to do a rollback for the given exceptions.

Is it an expected behaviour to not do a rollback for transactions leading to BadRequestDataException?

PR #74, opened 2020-06-18
feat: use message from the initial exception when available #41 #65

bobeal:


entity-service/src/test/kotlin/com/egm/stellio/entity/web/EntityHandlerTests.kt
```diff
@@ -195,6 +195,30 @@ class EntityHandlerTests {
             .expectStatus().isBadRequest
     }
 
+    @Test
+    fun `create entity should return a 400 if input data is not valid and creation was rejected`() {
+        val jsonLdFile = ClassPathResource("/ngsild/aquac/BreedingService.json")
+
+        every { entityService.exists(any()) } returns false
+        every { entityService.createEntity(any()) } throws BadRequestDataException("Target entity does not exist")
```
vraybaud, line 203: Shouldn't this be an exception with a cause to test it correctly works ?

bobeal: It aims to reproduce the behavior described in #41 and #65

vraybaud: Yes indeed but the behaviour was that the Exception was wrapped inside another Exception and this was not handled (according my understanding of the issues)

bobeal: I did not want to artificially create a wrapped exception but let it be created by the framework (as in normal operations and as described in the linked issues).

But indeed I realized the exception is received already unwrapped in the exception handler in this test config ...

vraybaud: > I did not want to artificially create a wrapped exception but let it be created by the framework (as in normal operations and as described in the linked issues).

This could be done in karate tests, right ?

bobeal: Yes but this is not totally satisfying, as there is still a difference of behavior at the web handler level between tests and runtime that I don't really understand ...

Anyway, in the meantime, I've modified the test to add a manual exception wrapping.

Issue #70, opened 2020-06-14
Improve queries in entity creation process


bobeal:
Current version does the job but is not very efficient in terms of queries made to the DB.

Improve that and also create some minimal indexes.

PR #75, opened 2020-06-18
feat(entity): improve queries in entity creation process #70

bobeal:
- create properties and relationships of an entity in one Cypher query
- add indexes on entities ids and properties names (the most commonly used)
- improve the efficiency of the entity existence query
- remove customs logs in Neo4jRepository as they can be activated at the Bolt driver level instead
- mark the @Transactional method at the service level as they usually group more than 1 query
- other minor code simplifications

### DIFF
--- a/entity-service/src/main/kotlin/com/egm/stellio/entity/repository/Neo4jRepository.kt
+++ b/entity-service/src/main/kotlin/com/egm/stellio/entity/repository/Neo4jRepository.kt
@@ -8,13 +8,12 @@ import com.egm.stellio.shared.util.isDate
 import com.egm.stellio.shared.util.isDateTime
 import com.egm.stellio.shared.util.isFloat
 import com.egm.stellio.shared.util.isTime
+import com.egm.stellio.shared.util.toRelationshipTypeName
 import org.neo4j.ogm.session.Session
 import org.neo4j.ogm.session.SessionFactory
 import org.neo4j.ogm.session.event.Event
 import org.neo4j.ogm.session.event.EventListenerAdapter
-import org.slf4j.LoggerFactory
 import org.springframework.stereotype.Component
-import org.springframework.transaction.annotation.Transactional
 import java.time.Instant
 import java.time.ZoneOffset
 import javax.annotation.PostConstruct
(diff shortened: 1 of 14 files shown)


entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityService.kt
```diff
@@ -79,33 +75,21 @@ class EntityService(
...
-        propertiesAndRelationshipsMap.filter { entry ->
-            isAttributeOfType(entry.value, NGSILD_RELATIONSHIP_TYPE)
-        }.forEach { entry ->
+        expandedEntity.relationships.forEach { entry ->
             val relationshipType = entry.key
             val objectId = getRelationshipObjectId(entry.value)
-            val objectEntity = entityRepository.findById(objectId)
 
-            if (!objectEntity.isPresent) {
+            if (!exists(objectId))
```
vraybaud, line 82: The validation could be done before. What I think is if the last Relationship is targeting a non existant entity, we did all the processing for the other relationships for nothing, what do you think ?

bobeal: Good idea. I guess we could even reuse part of the logic used in batch creation?

vraybaud: I thought about it but the code in batch creation is just checking against a list of entities given in parameter, I don't know if we can find a correlation and unify both.

entity-service/src/main/kotlin/com/egm/stellio/entity/repository/Neo4jRepository.kt
```diff
@@ -250,7 +251,7 @@ class Neo4jRepository(
                         it.third.isTime() -> "localtime('${it.third}')"
                         else -> "'${it.third}'"
                     }
-                    "(n)-[:HAS_VALUE]->(${it.first.extractShortTypeFromExpanded()}: Property { name: '${it.first}'}) and ${it.first.extractShortTypeFromExpanded()}.value ${it.second} $comparableValue"
+                    "(n)-[:HAS_VALUE]->(${it.first.extractShortTypeFromExpanded()}:Property { name: '${it.first}'}) and ${it.first.extractShortTypeFromExpanded()}.value ${it.second} $comparableValue"
```
vraybaud, line 254: This line is way too long

bobeal: Well, it was also before but I try to avoid mixing code formatting and feature / fix in a PR because that makes code reviews painful.

entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityService.kt
bobeal: It seems like I could just call `EntitiesGraphBuilder.build(...)` https://github.com/stellio-hub/stellio-context-broker/blob/develop/entity-service/src/main/kotlin/com/egm/stellio/entity/util/EntitiesGraphBuilder.kt#L26 with my single entity?

entity-service/src/main/kotlin/com/egm/stellio/entity/repository/Neo4jRepository.kt
```diff
@@ -24,30 +23,34 @@ class Neo4jRepository(
...
+            RETURN p.id as id
+        """
+
+        return session.query(query, mapOf("props" to property.nodeProperties()))
+            .first()["id"] as String
+    }
 
-        return session.query(query, emptyMap<String, String>()).queryStatistics().relationshipsCreated
+    fun createRelationshipOfSubject(subjectId: String, relationship: Relationship, targetId: String): String {
+        val relationshipType = relationship.type[0].toRelationshipTypeName()
```
vraybaud, line 39: ```.type[0]``` I think this should be hidden in a method in Relationship class

shared/src/main/kotlin/com/egm/stellio/shared/util/NgsiLdParsingUtils.kt
```diff
@@ -123,7 +123,6 @@ object NgsiLdParsingUtils {
             throw BadRequestDataException("Could not parse entity due to invalid json-ld payload")
 
         val expandedResult = JsonUtils.toPrettyString(expandedEntity[0])
-        logger.debug("Expanded entity is $expandedResult")
```
vraybaud, line 126: This is awesome. :1st_place_medal:

PR #80, opened 2020-06-19
fix(entity): set wait_for_neo4j.sh timeout as variable with a default value 100

HoucemKacem:


entity-service/build.gradle.kts
```diff
@@ -36,8 +36,9 @@ tasks.bootRun {
 
 jib.from.image = project.ext["jibFromImage"].toString()
 jib.to.image = "stellio/stellio-entity-service"
-jib.container.entrypoint = listOf("/bin/sh", "-c", "/database/wait-for-neo4j.sh neo4j:7687 -t 40 -- " +
+jib.container.entrypoint = listOf("/bin/sh", "-c", "/database/wait-for-neo4j.sh neo4j:7687 -t \$NEO4J_START_TIMEOUT -- " +
     "java " + project.ext["jibContainerJvmFlag"].toString() + " -cp /app/resources:/app/classes:/app/libs/* " + mainClass)
+jib.container.environment = mapOf("NEO4J_START_TIMEOUT" to "100")
```
bobeal, line 41: What about calling it `NEO4J_WAIT_TIMEOUT`?

HoucemKacem: renamed

PR #82, opened 2020-06-20
feat(entity): improve cypher queries for efficiency #81

bobeal:
- use a node label in every MATCH clause
- use params instead of literals in queries (allows caching of query planning by neo4j)
- use existential subqueries to avoid cartesian products in entities search query
- remove HAS_OBJECT relationship between a relationship node and its target entity
- reduce intermediate results when deleting an entity or attribute

### DIFF
--- a/entity-service/src/main/kotlin/com/egm/stellio/entity/repository/Neo4jRepository.kt
+++ b/entity-service/src/main/kotlin/com/egm/stellio/entity/repository/Neo4jRepository.kt
@@ -3,7 +3,6 @@ package com.egm.stellio.entity.repository
 import com.egm.stellio.entity.model.Entity
 import com.egm.stellio.entity.model.Property
 import com.egm.stellio.entity.model.Relationship
-import com.egm.stellio.shared.util.extractShortTypeFromExpanded
 import com.egm.stellio.shared.util.isDate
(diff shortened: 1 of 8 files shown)


entity-service/src/main/kotlin/com/egm/stellio/entity/repository/Neo4jRepository.kt
```diff
@@ -18,33 +17,45 @@ import java.time.Instant
 import java.time.ZoneOffset
 import javax.annotation.PostConstruct
 
+sealed class SubjectNodeInfo(val id: String, val label: String)
+class EntitySubjectNode(id: String) : SubjectNodeInfo(id, "Entity")
+class AttributeSubjectNode(id: String) : SubjectNodeInfo(id, "Attribute")
```
vraybaud, line 22: Why not push it further ? RelationshipSubjectNode / PropertySubjectNode ? If I understand correctly the more label you put the faster it goes ?

bobeal: Because there is currently no use-case for such specializations. Other queries known on which labels they apply and don't need to be parameterized on label name.

PR #86, opened 2020-06-23
feat(gateway): use environment variables to load routes uri #83

HoucemKacem:


api-gateway/src/main/kotlin/com/egm/stellio/apigateway/ApiGatewayApplication.kt
```diff
@@ -11,6 +12,14 @@ import org.springframework.context.annotation.Bean
 class ApiGatewayApplication(
     private val filterFactory: TokenRelayGatewayFilterFactory
 ) {
+    @Value("\${entity.service.url:entity-service}")
```
bobeal, line 15: I prefer using the prefix `application` for all our properties (that allows distinguishing them from Spring or libs properties)

For instance `application.entity-service.url`

Issue #84, opened 2020-06-23
extend size of columns type, attribute_name, attribute_value_type in PostgreSQL database


vraybaud:
ERROR : Error: SEVERITY_LOCALIZED=ERROR, SEVERITY_NON_LOCALIZED=ERROR, CODE=22001, MESSAGE=value too long for type character varying(64), FILE=varchar.c, LINE=624, ROUTINE=varchar

How to reproduce : 

```json
POST /ngsi-ld/v1/entities 
{
    "id": "urn:ngsi-ld:Consumer:Test1",
    "type": "Consumer",
    "waterConsumption": {
        "type": "Property",
        "value": 0,
        "unitCode": "LTR",
        "observedAt":"1970-01-01T00:00:00Z"
    },
    "minFlow": {
        "type": "Property",
        "value": 0,
        "unitCode": "MQS",
        "observedAt":"1970-01-01T00:00:00Z"
    },
    "maxFlow": {
        "type": "Property",
        "value": 0,
        "unitCode": "MQS",
        "observedAt":"1970-01-01T00:00:00Z"
    },
    "persistenceFlowDuration": {
        "type": "Property",
        "value": 0,
        "unitCode": "HUR",
        "observedAt":"1970-01-01T00:00:00Z"
    },
    "acquisitionStageFailure": {
        "type": "Property",
        "value": 0,
        "observedAt":"1970-01-01T00:00:00Z"
    },
    "alarmFlowPersistence": {
        "type": "Property",
        "value": 0,
        "observedAt":"1970-01-01T00:00:00Z"
    },
    "alarmInProgressAlarms": {
        "type": "Property",
        "value": 0,
        "observedAt":"1970-01-01T00:00:00Z"
    },
    "alarmMetrology": {
        "type": "Property",
        "value": 0,
        "observedAt":"1970-01-01T00:00:00Z"
    },
    "alarmStopsLeaks": {
        "type": "Property",
        "value": 0,
        "observedAt":"1970-01-01T00:00:00Z"
    },
    "alarmSystem": {
        "type": "Property",
        "value": 0,
        "observedAt":"1970-01-01T00:00:00Z"
    },
    "alarmTamper": {
        "type": "Property",
        "value": 0,
        "observedAt":"1970-01-01T00:00:00Z"
    },
    "alarmWaterQuality": {
        "type": "Property",
        "value": 0,
        "observedAt":"1970-01-01T00:00:00Z"
    },
    "moduleTampered": {
[...]

PR #87, opened 2020-06-23
fix(search): extend size of stellio_search db columns #84

vraybaud:
columns : type, attribute_value_type, attribute_name


bobeal:
PS : use the fix prefix (not bug) in your branch name (for the double tree already discussed and also for automatic labelling by PR labeler).

PR #102, opened 2020-07-06
feature(entity): add support for multi attribute in create/retrieve entity

HoucemKacem:
Implementation for support for multi-attributes:
-  Create and retrieve an entity (entity service)


entity-service/src/main/kotlin/com/egm/stellio/entity/model/Attribute.kt
```diff
@@ -28,6 +28,8 @@ open class Attribute(
     @JsonIgnore
     var modifiedAt: ZonedDateTime? = null,
 
+    var datasetId: String? = null,
```
bobeal, line 31: `datasetId` is supposed to be of URI type. It is a good time to start using this (we should have done it for id). Does neo4j have a type for URI?

bobeal:
You have a bunch of continuation indents that are at 8 instead of 4, please check them.

entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityService.kt
```diff
@@ -181,27 +190,34 @@ class EntityService(
...
-            // for short-handed properties, the value is directly accessible from the map under the @value key
-            val propertyValue =
-                getPropertyValueFromMap(entry.value, NGSILD_PROPERTY_VALUE) ?: entry.value["@value"]!!
-
-            val rawProperty = Property(
-                name = entry.key, value = propertyValue,
-                unitCode = getPropertyValueFromMapAsString(entry.value, NGSILD_UNIT_CODE_PROPERTY),
-                observedAt = getPropertyValueFromMapAsDateTime(entry.value, NGSILD_OBSERVED_AT_PROPERTY)
-            )
+            entry.value.forEach { instance ->
```
bobeal, line 198: As I understand the spec, multi-attribute applies to entities, not to attributes.

entity-service/src/main/kotlin/com/egm/stellio/entity/model/Attribute.kt
HoucemKacem: As I see [here](https://neo4j.com/docs/cypher-manual/current/syntax/values/), there is no support for URI data type.
Seeing the spatial type Point makes me wonder about the other spatial type that we could have.

entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityService.kt
HoucemKacem: Are you referring to that ? 
"For each Entity, there can be Attributes that simultaneously have more than one instance."

bobeal: Indeed

entity-service/src/main/kotlin/com/egm/stellio/entity/model/Attribute.kt
bobeal: Ok, so we could have it as an URI in the code and persisted as a string?

HoucemKacem: Yes

entity-service/src/main/kotlin/com/egm/stellio/entity/model/Attribute.kt
```diff
@@ -60,6 +62,11 @@ open class Attribute(
             )
         }
 
+        datasetId?.run {
+            if (this != "default")
```
vraybaud, line 66: "default" should be a constant, so if this is changing the modification is straightforward

entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityService.kt
```diff
@@ -131,12 +135,17 @@ class EntityService(
         val propertyValue = getPropertyValueFromMap(propertyValues, NGSILD_PROPERTY_VALUE)
             ?: throw BadRequestDataException("Key $NGSILD_PROPERTY_VALUE not found in $propertyValues")
 
+        val datasetId = getPropertyValueFromMapAsString(propertyValues, NGSILD_DATASET_ID_PROPERTY)
+        if (datasetId == null && neo4jRepository.hasPropertyWithDefaultInstance(EntitySubjectNode(entity.id), propertyKey))
+            throw BadRequestDataException("Property $propertyKey already has default instance")
```
vraybaud, line 140: To optimize performances, we should think about a solution in which you check that the payload is valid (meaning only one default attribute, unique datasetIds) before launching the creation. What do you think @HoucemKacem @bobeal ?

vraybaud: Also, it creates less noises in the already big EntityService.

entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityService.kt
```diff
@@ -268,61 +278,61 @@ class EntityService(
...
-            .groupBy {
-                (it["property"] as Property).id
-            }
-            .values
-            .forEach {
-                val (propertyKey, propertyValues) = buildPropertyFragment(it, entity.contexts)
-                resultEntity[propertyKey] = propertyValues
-            }
+        .groupBy {
+            (it["property"] as Property).id
```
vraybaud, line 286: Will this really work ? I'm afraid if there are multiple properties returned by getEntitySpecificProperties this will be an error, right ?

HoucemKacem: Actually the properties will not have the same id.
I don't see how it could raise an error. 
@bobeal, any ideas ?

entity-service/src/main/kotlin/com/egm/stellio/entity/model/Attribute.kt
```diff
@@ -60,6 +65,11 @@ open class Attribute(
             )
         }
 
+        datasetId?.run {
+            if (this.toString() != "default")
```
vraybaud, line 69: "default" should be a constant

entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityService.kt
```diff
@@ -268,61 +278,61 @@ class EntityService(
...
+        }
+        .values
+        .forEach {
```
vraybaud, line 289:
```
.map {  buildInstanceFragment(it, entity.contexts) }
.groupBy { it._1 }
forEach { resultEntity[it._1] = it._2 }
```
Doesn't this work and avoid to make resultEntity a var ?

entity-service/src/main/kotlin/com/egm/stellio/entity/repository/Neo4jRepository.kt
```diff
@@ -425,6 +425,19 @@ class Neo4jRepository(
             .firstOrNull()
     }
 
+    fun hasPropertyWithDefaultInstance(subjectNodeInfo: SubjectNodeInfo, propertyName: String): Boolean {
+        val query = """
+            MATCH (a:${subjectNodeInfo.label} { id: ${'$'}attributeId })-[:HAS_VALUE]->(property:Property { name: ${'$'}propertyName, datasetId: "default" })
```
vraybaud, line 430: use constant for default

entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityService.kt
HoucemKacem: I think it should be: 
```
.map {
    buildInstanceFragment(it, entity.contexts)
}
.groupBy { it.first }
.forEach {
    it.value.forEach {
    resultEntity[it.first] = it.second
    }
}
```
I tried this but it keeps only the last instance in the resultEntity.

entity-service/src/main/kotlin/com/egm/stellio/entity/repository/Neo4jRepository.kt
```diff
@@ -3,11 +3,7 @@ package com.egm.stellio.entity.repository
 import com.egm.stellio.entity.model.Entity
 import com.egm.stellio.entity.model.Property
 import com.egm.stellio.entity.model.Relationship
-import com.egm.stellio.shared.util.isDate
```
bobeal, line 6: you only have diffs in import in this file, remove them

entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityService.kt
```diff
@@ -96,7 +100,12 @@ class EntityService(
         }
 
         expandedEntity.properties.forEach { entry ->
-            createEntityProperty(entity, entry.key, entry.value)
+            if (entry.value.count { !it.containsKey(NGSILD_DATASET_ID_PROPERTY) } > 1)
+                throw BadRequestDataException("Property ${entry.key.extractShortTypeFromExpanded()} already has default instance")
```
bobeal, line 104: The message is not clear about the error. It should instead say that you can't have more than one default instance.

entity-service/src/test/kotlin/com/egm/stellio/entity/web/EntityHandlerTests.kt
```diff
@@ -346,6 +349,68 @@ class EntityHandlerTests {
...
+                ),
+                listOf(NGSILD_CORE_CONTEXT)
+        )
+
+        webClient.get()
+                .uri("/ngsi-ld/v1/entities/urn:ngsi-ld:BeeHive:TESTC")
+                .header(HttpHeaders.CONTENT_TYPE, MediaType.APPLICATION_JSON_VALUE)
+                .exchange()
+                .expectStatus().isOk
+                .expectBody().json("{\"id\":\"urn:ngsi-ld:Beehive:4567\",\"type\":\"Beehive\",\"name\":{\"type\":\"Property\",\"datasetId\":\"urn:ngsi-ld:Property:french-name\",\"value\":\"ruche\"},\"@context\":\"$NGSILD_CORE_CONTEXT\"}")
```
bobeal, line 376: switch to multiline string, it is easier to read

entity-service/src/test/kotlin/com/egm/stellio/entity/web/EntityHandlerTests.kt
```diff
@@ -346,6 +349,68 @@ class EntityHandlerTests {
...
+                ),
+                listOf(NGSILD_CORE_CONTEXT)
+        )
+
+        webClient.get()
+                .uri("/ngsi-ld/v1/entities/urn:ngsi-ld:BeeHive:TESTC")
+                .header(HttpHeaders.CONTENT_TYPE, MediaType.APPLICATION_JSON_VALUE)
+                .exchange()
+                .expectStatus().isOk
+                .expectBody().json("{\"id\":\"urn:ngsi-ld:Beehive:4567\",\"type\":\"Beehive\",\"name\":[{\"type\":\"Property\",\"datasetId\":\"urn:ngsi-ld:Property:english-name\",\"value\":\"beehive\"},{\"type\":\"Property\",\"datasetId\":\"urn:ngsi-ld:Property:french-name\",\"value\":\"ruche\"}],\"@context\":\"$NGSILD_CORE_CONTEXT\"}")
```
bobeal, line 411: switch to multiline string, it is easier to read

entity-service/src/test/resources/ngsild/aquac/BreedingService_propWithMoreThanDefaultInstance.json
```diff
@@ -0,0 +1,19 @@
+{
```
bobeal, line 1: is it used?

entity-service/src/test/kotlin/com/egm/stellio/entity/service/EntityServiceTests.kt
```diff
@@ -548,6 +580,99 @@ class EntityServiceTests {
         confirmVerified()
     }
 
+    @Test
+    fun `it should create a property instance with a default datasetId`() {
```
bobeal, line 584: I think those 2 tests do not bring a lot. Basically you are just testing the constructor of a Property object ...

entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityService.kt
```diff
@@ -136,7 +145,8 @@ class EntityService(
         val rawProperty = Property(
             name = propertyKey, value = propertyValue,
             unitCode = getPropertyValueFromMapAsString(propertyValues, NGSILD_UNIT_CODE_PROPERTY),
-            observedAt = getPropertyValueFromMapAsDateTime(propertyValues, NGSILD_OBSERVED_AT_PROPERTY)
+            observedAt = getPropertyValueFromMapAsDateTime(propertyValues, NGSILD_OBSERVED_AT_PROPERTY),
+            datasetId = getPropertyValueFromMapAsUri(propertyValues, NGSILD_DATASET_ID_PROPERTY) ?: URI.create(NGSILD_DATASET_ID_DEFAULT_VALUE)
```
bobeal, line 149: I guess that implies every Property will have a datasetId? Is it what we want?

shared/src/main/kotlin/com/egm/stellio/shared/util/NgsiLdParsingUtils.kt
```diff
@@ -194,8 +199,10 @@ object NgsiLdParsingUtils {
                         NGSILD_TIME_TYPE -> LocalTime.parse(finalValue)
                         else -> firstListEntry["@value"]
                     }
-                } else {
+                } else if (firstListEntry["@value"] != null) {
                     firstListEntry["@value"]
+                } else {
```
bobeal, line 204: why did you have to do that?

shared/src/main/kotlin/com/egm/stellio/shared/util/NgsiLdParsingUtils.kt
```diff
@@ -246,6 +258,15 @@ object NgsiLdParsingUtils {
             values[NGSILD_ENTITY_TYPE] is List<*> &&
             values.getOrElse(NGSILD_ENTITY_TYPE) { emptyList() }[0] == type.uri
 
+    /**
+     * Given an entity's attribute, returns whether all its instances are of the given attribute type
+     * (i.e. property)
+     */
+    fun isAttributeOfType(values: List<Map<String, List<Any>>>, type: AttributeType): Boolean =
```
bobeal, line 265: This is a little bit weird. What if a multi-attribute is a mix of properties and relationships? (yes, it should be rejected)

entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityService.kt
```diff
@@ -276,9 +286,15 @@ class EntityService(
                 (it["property"] as Property).id
             }
             .values
+            .map { buildInstanceFragment(it, entity.contexts) }
+            .groupBy { it.first }
```
vraybaud, line 290: To think of it again, I think it would be fine to replace it by variables here to improve readibility... What do you think ?

entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityService.kt
vraybaud: To me it's still the same, you do processing but you could know that this would fail before just by cheking the payload.

entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityService.kt
HoucemKacem: You mean for non multi-attribute cases the datasetId should be null ?

shared/src/main/kotlin/com/egm/stellio/shared/util/NgsiLdParsingUtils.kt
HoucemKacem: Since it is expanded with the following format:
```
{ ......
    "https://uri.etsi.org/ngsi-ld/datasetId" = [{
        "@id"= dataset_id_value
    }]
 ......
}
```

shared/src/main/kotlin/com/egm/stellio/shared/util/NgsiLdParsingUtils.kt
HoucemKacem: indeed

HoucemKacem: Here we are checking if all attribute instances are of type property.
If not we return false and the attribute will be rejected.

entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityService.kt
HoucemKacem: Yes

entity-service/src/test/kotlin/com/egm/stellio/entity/web/EntityHandlerTests.kt
```diff
@@ -346,6 +349,77 @@ class EntityHandlerTests {
...
+                ),
+                listOf(NGSILD_CORE_CONTEXT)
+        )
+
+        webClient.get()
+                .uri("/ngsi-ld/v1/entities/urn:ngsi-ld:BeeHive:TESTC")
+                .header(HttpHeaders.CONTENT_TYPE, MediaType.APPLICATION_JSON_VALUE)
+                .exchange()
+                .expectStatus().isOk
+                .expectBody().json(
```
bobeal, line 376: I meant using a Kotlin multi-line string, not to add carriage returns (which is still bloated)!

shared/src/main/kotlin/com/egm/stellio/shared/util/NgsiLdParsingUtils.kt
bobeal: rejected or silently ignored?

HoucemKacem: By rejected you mean an exception should be raised and the entity cannot be created ?

shared/src/main/kotlin/com/egm/stellio/shared/model/ExpandedEntity.kt
```diff
@@ -102,4 +104,21 @@ class ExpandedEntity private constructor(
...
     }
+
+    fun propertiesHaveAtMostOneDefaultInstance(): Boolean {
+        return this.properties.all { property ->
+            property.value.count { !it.containsKey(NGSILD_DATASET_ID_PROPERTY) } < 2
+        }
+    }
+
+    fun propertiesHaveNoDuplicatedDatasetId(): Boolean {
+        return this.properties.all { property ->
```
vraybaud, line 115: "this" isn't required in this scope

shared/src/main/kotlin/com/egm/stellio/shared/model/ExpandedEntity.kt
```diff
@@ -80,4 +89,36 @@ class ExpandedEntity private constructor(
...
+            val datasetIds = property.value.map {
+                val datasetId = it[NGSILD_DATASET_ID_PROPERTY]?.get(0) as Map<String, String>?
+                datasetId?.get(NGSILD_ENTITY_ID)
+            }
+
+            datasetIds.distinct().count() == datasetIds.count()
```
vraybaud, line 121: datasetIds.toSet.count() == datasetIds.count()?

entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityService.kt
```diff
@@ -86,6 +90,12 @@ class EntityService(
             throw BadRequestDataException("Entity ${expandedEntity.id} targets unknown entities: $inErrorRelationships")
         }
 
+        if (! expandedEntity.propertiesHaveAtMostOneDefaultInstance())
```
vraybaud, line 93: Strange space between ! and expandedEntity

entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityService.kt
```diff
@@ -86,6 +90,12 @@ class EntityService(
             throw BadRequestDataException("Entity ${expandedEntity.id} targets unknown entities: $inErrorRelationships")
         }
 
+        if (! expandedEntity.propertiesHaveAtMostOneDefaultInstance())
+            throw BadRequestDataException("Properties can't have more than one default instance")
+
+        if (! expandedEntity.propertiesHaveNoDuplicatedDatasetId())
```
vraybaud, line 96: ```propertiesHaveUniqueDatasetId```
naming should not be negative when it's possible

shared/src/main/kotlin/com/egm/stellio/shared/util/NgsiLdParsingUtils.kt
bobeal: add it as a comment in the code

entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityService.kt
```diff
@@ -86,6 +88,12 @@ class EntityService(
             throw BadRequestDataException("Entity ${expandedEntity.id} targets unknown entities: $inErrorRelationships")
         }
 
+        if (!expandedEntity.propertiesHaveAtMostOneDefaultInstance())
```
vraybaud, line 91: Maybe not in this jira but I think it's time to put all of this validation in a separate component

bobeal: Jira 😨 (please, no).

PS: open an issue for this validation topic.

bobeal: btw, there is already some validation set up in ExpandedEntity, it might be a better place.

bobeal: what bothers me here is that we don't say to the user which property has the error

entity-service/src/test/kotlin/com/egm/stellio/entity/service/EntityServiceTests.kt
```diff
@@ -233,6 +233,93 @@ class EntityServiceTests {
...
+        every { neo4jRepository.createPropertyOfSubject(any(), any()) } returns UUID.randomUUID().toString()
+        every { repositoryEventsListener.handleRepositoryEvent(any()) } just Runs
+        every { entityRepository.getEntityCoreById(any()) } returns mockedBreedingService
+        every { mockedBreedingService.serializeCoreProperties() } returns mutableMapOf(
+            "@id" to "urn:ngsi-ld:BreedingService:PropWithMoreThanOneDefaultInstance",
+            "@type" to listOf("BreedingService")
+        )
+        every { entityRepository.getEntitySpecificProperties(any()) } returns listOf()
+        every { entityRepository.getEntityRelationships(any()) } returns listOf()
+        every { mockedBreedingService.contexts } returns sampleDataWithContext.contexts
```
bobeal, line 256: All those are not actually called and can be removed

bobeal: And same for the 2 other tests.

shared/src/main/kotlin/com/egm/stellio/shared/model/ExpandedEntity.kt
```diff
@@ -24,21 +26,32 @@ class ExpandedEntity private constructor(
...
-            NgsiLdParsingUtils.expandValueAsMap(it.value)
-        }.filter {
-            NgsiLdParsingUtils.isAttributeOfType(it.value, type)
+    private fun getAttributesOfType(type: AttributeType): Any =
+        when (type) {
+            NGSILD_PROPERTY_TYPE -> attributes.mapValues {
+                NgsiLdParsingUtils.expandValueAsListOfMap(it.value)
+            }
+            .filter {
+                NgsiLdParsingUtils.isValidAttribute(it.value)
```
bobeal, line 45: that can also be part of early validation

shared/src/main/kotlin/com/egm/stellio/shared/model/ExpandedEntity.kt
```diff
@@ -109,20 +110,22 @@ class ExpandedEntity private constructor(
...
+        properties.forEach { property ->
+            if (property.value.count { !it.containsKey(NGSILD_DATASET_ID_PROPERTY) } > 1)
+                throw BadRequestDataException("Property ${property.key.extractShortTypeFromExpanded()} can't have more than one default instance")
         }
     }
 
-    fun propertiesHaveUniqueDatasetId(): Boolean {
-        return properties.all { property ->
+    fun checkPropertiesHaveUniqueDatasetId() {
+        properties.forEach { property ->
```
vraybaud, line 121: ``` property.find { ... } ?.let { throw ... } ```

Could this work ? (just a different style)

HoucemKacem: property.first ?

vraybaud: properties I meant

vraybaud: I thought properties.first { } returned T? but actually it throws an exception if no element is found... what works is properties.find { } which actually returns a T?
so it should be :
```properties.find { ... } ?. let { throw ... }

see : https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/find.html & https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/first.html

bobeal: There isn't a `find` method on maps.

Current version may probably be slightly improved but yet it is simple and readable. So let's stick with it.

PR #105, opened 2020-07-09
feat: expandedEntity validation

HoucemKacem:


HoucemKacem:
When trying to create an entity it returns the following error:
```
{
    "detail": "class java.util.LinkedHashMap cannot be cast to class java.util.List (java.util.LinkedHashMap and java.util.List are in module java.base of loader 'bootstrap')",
    "title": "There has been an error during the operation execution",
    "type": "https://uri.etsi.org/ngsi-ld/errors/InternalError"
}

```

HoucemKacem:
There were a second call to getAttributesOfType() declared in the companion object.
With attributes that contains temporalValues.

PR #110, opened 2020-07-10
feat(entity): add support for multi-attribute in update entity attributes

HoucemKacem:


entity-service/src/main/kotlin/com/egm/stellio/entity/repository/Neo4jRepository.kt
```diff
@@ -134,6 +135,28 @@ class Neo4jRepository(
         return session.query(query, parameters, true).toList().isNotEmpty()
     }
 
+    fun hasPropertyInstance(subjectNodeInfo: SubjectNodeInfo, propertyName: String, datasetId: String?): Boolean {
+        val query = if (datasetId == null)
+            """
+            MATCH (a:${subjectNodeInfo.label} { id: ${'$'}attributeId })-[:HAS_VALUE]->(property:Property { name: ${'$'}propertyName })
+            WHERE NOT EXISTS (property.datasetId)
```
vraybaud, line 142: How do you access default property ? This is for the old properties ? Shouldn't we migrate what's already in the DB ?

HoucemKacem: The default property is the property having no datasetId (we can say that all the existing properties in the database are default).
To update the default property the update fragment should not contain a datasetId.
Yes this works for the old properties.
No data migration is needed.

entity-service/src/main/kotlin/com/egm/stellio/entity/repository/Neo4jRepository.kt
```diff
@@ -393,15 +416,22 @@ class Neo4jRepository(
         return session.query(query, mapOf("entitiesIds" to entitiesIds), true).map { it["id"] as String }
     }
 
-    fun getPropertyOfSubject(subjectId: String, propertyName: String): Property {
-        val query = """
+    fun getPropertyOfSubject(subjectId: String, propertyName: String, datasetId: URI? = null): Property? {
```
bobeal, line 419: I'm not really happy with returned Property being nullable, it forces to make (always dangerous) not null assertions

entity-service/src/test/kotlin/com/egm/stellio/entity/repository/Neo4jRepositoryTests.kt
```diff
@@ -533,6 +534,54 @@ class Neo4jRepositoryTests {
         neo4jRepository.deleteEntity(device.id)
     }
 
+    @Test
+    fun `it should return the default property instance`() {
```
bobeal, line 538:
```suggestion
    fun `it should return the default property instance if no datasetId is provided`() {
```

entity-service/src/test/kotlin/com/egm/stellio/entity/repository/Neo4jRepositoryTests.kt
```diff
@@ -533,6 +534,54 @@ class Neo4jRepositoryTests {
...
+            listOf("Beekeeper"),
+            mutableListOf(Property(name = "name", value = 100L))
+        )
+
+        assertNotNull(neo4jRepository.getPropertyOfSubject(entity.id, "name", null))
+        neo4jRepository.deleteEntity(entity.id)
+    }
+
+    @Test
+    fun `it should not return a default property instance`() {
```
bobeal, line 550:
```suggestion
    fun `it should not return a default property instance if a datasetId is provided`() {
```

entity-service/src/test/kotlin/com/egm/stellio/entity/repository/Neo4jRepositoryTests.kt
```diff
@@ -533,6 +534,54 @@ class Neo4jRepositoryTests {
...
+            listOf("Beekeeper"),
+            mutableListOf(Property(name = "name", value = 100L, datasetId = URI.create("urn:ngsi-ld:Dataset:name:1")))
+        )
+
+        assertNotNull(neo4jRepository.getPropertyOfSubject(entity.id, "name", datasetId = URI.create("urn:ngsi-ld:Dataset:name:1")))
+        neo4jRepository.deleteEntity(entity.id)
+    }
+
+    @Test
+    fun `it should not return the property instance with the given datasetId`() {
```
bobeal, line 574:
```suggestion
    fun `it should not return the property instance if the given datasetId does not match`() {
```

entity-service/src/main/kotlin/com/egm/stellio/entity/repository/Neo4jRepository.kt
HoucemKacem: I will fix this

entity-service/src/test/kotlin/com/egm/stellio/entity/service/EntityServiceTests.kt
```diff
@@ -456,14 +457,56 @@ class EntityServiceTests {
...
+                "unitCode": "months",
+                "datasetId": "urn:ngsi-ld:Dataset:fishAge:1"
+              }
+            }
+        """.trimIndent()
+
+        val mockkedSensor = mockkClass(Entity::class)
+        val mockkedPropertyEntity = mockkClass(Property::class)
+
+        every { mockkedSensor.id } returns sensorId
```
bobeal, line 486: maybe relaxed mockk(s) could save some useless mockked calls below?

entity-service/src/main/kotlin/com/egm/stellio/entity/repository/Neo4jRepository.kt
bobeal: In the previous method, datasetId parameter is a string. In this one, it is an URI. Better be consistent.

bobeal:
Also add a few tests for the new method in `neo4jRepository`

Issue #101, opened 2020-07-06
Add support for multi-attribute (4.5.5)


HoucemKacem:
Support for multi-attributes should be added as follow:

- [x] Create and retrieve an entity Properties (entity service)
- [x] Update entity Properties (entity service)
- [x] Store distinct entries in search service (search service)
- [x] Create and retrieve an entity Relationships (entity service)
- [x] Append entity Properties (entity service)
- [x] Search for the temporal evolution of a given instance (search service)
- [x] Delete entity Properties (entity attributes)
- [x] Define the event model
- [x] Update entity Relationships (entity service)
- [x] Append entity Relationships (entity service)
- [x] Delete entity Relationships (entity attributes)
- [x] Partial update entity Attributes (entity service)
- [ ] Integration in subscription service for Properties (subscription service)
- [ ] Integration in subscription service for Relationships (subscription service)
- [ ] Create and retrieve an entity GeoProperties (entity service)
- [ ] Update entity GeoProperties (entity service)
- [ ] Append entity GeoProperties (entity service)
- [ ] Delete entity GeoProperties (entity attributes)
- [ ] Integration in subscription service for GeoProperties (subscription service)
- [ ] Integration in search service for GeoProperties (Search service)

HoucemKacem:
@bobeal can we discuss the priorities ?

bobeal:
Go for next iteration: support for relationships.

PR #111, opened 2020-07-11
feat(search): integrate creation and update of a multi attribute property #101

bobeal:


search-service/src/main/kotlin/com/egm/stellio/search/listener/EntityListener.kt
```diff
@@ -41,24 +41,31 @@ class EntityListener(
...
-                //  - only add an attribute instance if observedAt (?)
                 //  - needs optimization (lot of JSON-LD parsing, ...)
                 val rawParsedData = jacksonObjectMapper().readTree(entityEvent.payload!!)
                 val rawEntity = parseEntity(entityEvent.updatedEntity!!)
                 val attributeName = rawParsedData.fieldNames().next()
-                val rawAttributeValue = rawParsedData[attributeName]["value"]
+                val attributeValuesNode = rawParsedData[attributeName]
+
+                if (!attributeValuesNode.has("observedAt")) {
+                    logger.debug("Ignoring update event for $attributeName, it has no observedAt information")
```
vraybaud, line 51: Should be at least info or warning as it has functional value ? (any user could be interested in that part, right?)

bobeal: Yes, INFO level makes sense (not WARNING, because it is not an unexpected scenario)

search-service/src/main/kotlin/com/egm/stellio/search/service/TemporalEntityAttributeService.kt
```diff
@@ -67,37 +69,33 @@ class TemporalEntityAttributeService(
...
+                it.value.map { instance ->
+                    Pair(it.key, instance)
                 }
-            }
+            })
 
         return Flux.fromIterable(temporalProperties.asIterable())
             .map {
-                val expandedValues = expandValueAsMap(it.value)
+                val expandedValues = it.second
```
vraybaud, line 86: you could do ```(name, expandedValues) ->``` (save one line of code, save the planet)

PR #112, opened 2020-07-13
feat(entity): add support for multi attribute relationships #101

HoucemKacem:


shared/src/main/kotlin/com/egm/stellio/shared/model/ExpandedEntity.kt
```diff
@@ -72,26 +73,29 @@ class ExpandedEntity private constructor(
...
-                .filter {
-                    NgsiLdParsingUtils.isAttributeOfType(it, NGSILD_RELATIONSHIP_TYPE)
-                }.map {
-                    NgsiLdParsingUtils.getRelationshipObjectId(it)
-                }
+                    .filter {
+                        NgsiLdParsingUtils.isAttributeOfType(it, NGSILD_RELATIONSHIP_TYPE)
+                    }.map {
+                        NgsiLdParsingUtils.getRelationshipObjectId(it)
+                    }
```
bobeal, line 97:
```suggestion
                .filter {
                    NgsiLdParsingUtils.isAttributeOfType(it, NGSILD_RELATIONSHIP_TYPE)
                }.map {
                    NgsiLdParsingUtils.getRelationshipObjectId(it)
                }
```

shared/src/main/kotlin/com/egm/stellio/shared/model/ExpandedEntity.kt
```diff
@@ -110,22 +114,41 @@ class ExpandedEntity private constructor(
         }
     }
 
-    fun checkPropertiesHaveAtMostOneDefaultInstance() {
-        properties.forEach { property ->
-            if (property.value.count { !it.containsKey(NGSILD_DATASET_ID_PROPERTY) } > 1)
-                throw BadRequestDataException("Property ${property.key.extractShortTypeFromExpanded()} can't have more than one default instance")
+    fun checkAttributesHaveAtMostOneDefaultInstance() {
+        checkAttributesHaveAtMostOneDefaultInstance(NGSILD_PROPERTY_TYPE)
```
vraybaud, line 118: Best would be to go with something like : 
```kotlin
fun checkAttributesHaveAtMostOneDefaultInstance() {
	properties.checkAttributesHaveAtMostOneDefaultInstance()?.let { throw BadRequestDataException("Property ${it.extractShortTypeFromExpanded()} can't have more than one default instance") }
	relationships.checkAttributesHaveAtMostOneDefaultInstance()?.let { throw BadRequestDataException("Relationship ${it.extractShortTypeFromExpanded()} can't have more than one default instance") }
}

private fun Map<String, List<Map<String, List<Any>>>>.checkAttributesHaveAtMostOneDefaultInstance(): String? {
    this.forEach { attribute ->
        if (attribute.value.count { !it.containsKey(NGSILD_DATASET_ID_PROPERTY) } > 1) {
        	return attribute.key
        }
    }
}
```
In my opinion it's better as the shared method stays neutral whether the type is.

vraybaud: And I really like the Kotlin extensions

shared/src/main/kotlin/com/egm/stellio/shared/model/ExpandedEntity.kt
```diff
@@ -110,22 +114,33 @@ class ExpandedEntity private constructor(
...
+        properties.checkAttributesHaveAtMostOneDefaultInstance()?.let { throw BadRequestDataException("Property ${it.extractShortTypeFromExpanded()} can't have more than one default instance") }
+        relationships.checkAttributesHaveAtMostOneDefaultInstance()?.let { throw BadRequestDataException("Relationship ${it.extractShortTypeFromExpanded()} can't have more than one default instance") }
+    }
+
+    fun checkAttributesHaveUniqueDatasetId() {
+        properties.checkAttributesHaveUniqueDatasetId()?.let { throw BadRequestDataException("Property ${it.extractShortTypeFromExpanded()} can't have duplicated datasetId") }
+        relationships.checkAttributesHaveUniqueDatasetId()?.let { throw BadRequestDataException("Relationship ${it.extractShortTypeFromExpanded()} can't have duplicated datasetId") }
+    }
+
+    private fun Map<String, List<Map<String, List<Any>>>>.checkAttributesHaveAtMostOneDefaultInstance(): String? {
```
bobeal, line 127: Not really fan of the method contract.

bobeal: We have to guess that a method called `checkSomething` returns `null` if the check has failed

shared/src/main/kotlin/com/egm/stellio/shared/model/ExpandedEntity.kt
```diff
@@ -115,40 +115,32 @@ class ExpandedEntity private constructor(
...
-        val attributes = if (type == NGSILD_PROPERTY_TYPE) properties else relationships
-        attributes.forEach { attribute ->
-            if (attribute.value.count { !it.containsKey(NGSILD_DATASET_ID_PROPERTY) } > 1) {
-                if (type == NGSILD_PROPERTY_TYPE)
-                    throw BadRequestDataException("Property ${attribute.key.extractShortTypeFromExpanded()} can't have more than one default instance")
-                else
-                    throw BadRequestDataException("Relationship ${attribute.key.extractShortTypeFromExpanded()} can't have more than one default instance")
-            }
+    private fun Map<String, List<Map<String, List<Any>>>>.checkAttributesHaveAtMostOneDefaultInstance(): String? {
+        this.forEach { attribute ->
```
vraybaud, line 128: this.find { ... } could be used to avoid the return null

shared/src/main/kotlin/com/egm/stellio/shared/model/ExpandedEntity.kt
vraybaud: findAttributeWithMoreThanOneDefaultInstance ?

HoucemKacem: findAttributeWithMoreThanOneDefaultInstance is more meaningful.

shared/src/main/kotlin/com/egm/stellio/shared/model/ExpandedEntity.kt
HoucemKacem: there is no find method that can be applied to this.

vraybaud: yes indeed it would require to convert it to a list before...

shared/src/main/kotlin/com/egm/stellio/shared/model/ExpandedEntity.kt
bobeal: go for this renaming and close this PR.

PR #117, opened 2020-07-14
feat(WIP): PoC for a new implementation for entity validation

bobeal:
- use Arrow validated for validating an input JSON-LD payload
- currently only validate id, type and multi-attributes
- separate the JSON-LD expanded entity representation from the "parsed" NGSI-LD entity
- use JsonLdExpandedEntity as intermediate structure before returning an entity to a consumer


bobeal:
Just a PoC to launch the discussion :)

vraybaud:
So for now you only moved the validation linked to the payload format (multiple datasetIds, ...), we agree that the validation of the Expanded Entity against the DB (linked entities valid, entity already existing) should also be refactored ?

shared/src/main/kotlin/com/egm/stellio/shared/model/ExpandedEntity.kt
```diff
@@ -111,23 +110,84 @@ class ExpandedEntity private constructor(
...
-                datasetId?.get(NGSILD_ENTITY_ID)
-            }
-
-            if (datasetIds.toSet().count() != datasetIds.count())
-                throw BadRequestDataException("Property ${property.key.extractShortTypeFromExpanded()} can't have duplicated datasetId")
+fun checkProperties(rawJsonLdProperties: Map<String, Any>): Validated<Nel<EntityParsingError>, Map<String, PropertyInstances>> {
+    val rawProperties = getAttributesOfType(rawJsonLdProperties, NGSILD_PROPERTY_TYPE) as Map<String, PropertyInstances>
+    val errors = arrayListOf<EntityParsingError>()
+    rawProperties.forEach { (key, values) ->
+        if (!isValidAttribute(key, values))
```
vraybaud, line 145: isValidAttribute is not really a good name for a method, is it ?

shared/src/main/kotlin/com/egm/stellio/shared/model/ExpandedEntity.kt
```diff
@@ -1,68 +1,67 @@
...
 import com.egm.stellio.shared.util.extractShortTypeFromExpanded
 import com.github.jsonldjava.core.JsonLdOptions
 import com.github.jsonldjava.core.JsonLdProcessor
 
-class ExpandedEntity private constructor(
+// TODO we define a list of errors, this may not be necessary as we ultimately return the reason string to the consumers
+sealed class EntityParsingError(val reason: String) {
+    object MissingIdError : EntityParsingError("The provided NGSI-LD entity does not contain an id property")
+    object MissingTypeError : EntityParsingError("The provided NGSI-LD entity does not contain a type property")
+    class MixedAttributeInstancesError(reason: String) : EntityParsingError(reason)
```
vraybaud, line 33: maybe only the attribute name could be passed as a parameter there, no ?

```
class MixedAttributeInstancesError(attributeName: String) : EntityParsingError("$attributeName attribute instances must have the same type")
```

shared/src/main/kotlin/com/egm/stellio/shared/model/ExpandedEntity.kt
```diff
@@ -1,68 +1,67 @@
...
-                throw BadRequestDataException("The provided NGSI-LD entity does not contain an id property")
-
-            if (!attributes.containsKey("@type"))
-                throw BadRequestDataException("The provided NGSI-LD entity does not contain a type property")
-
-            return ExpandedEntity(attributes, contexts)
-        }
-    }
-
+    // FIXME kinda hacky but we often just need the id ... how can it be improved?
```
vraybaud, line 44: Move the part in which we need the id after the conversion to an expanded entity could be a solution, no?

shared/src/main/kotlin/com/egm/stellio/shared/model/ExpandedEntity.kt
```diff
@@ -1,68 +1,67 @@
...
-            .filter {
-                NgsiLdParsingUtils.isAttributeOfType(it.value, type)
-            }
-            else -> attributes.mapValues {
-                NgsiLdParsingUtils.expandValueAsMap(it.value)
-            }.filter {
-                NgsiLdParsingUtils.isAttributeOfType(it.value, type)
-            }
-        }
+typealias PropertyInstances = List<Map<String, List<Any>>>
```
vraybaud, line 51: amen

PR #120, opened 2020-07-16
feat(common): create Kubernetes config files #119

vraybaud:


bobeal:
@vraybaud what about this PR? reviewable / mergeable?

vraybaud:
> @vraybaud what about this PR? reviewable / mergeable?

This could be reviewable as it is working with minikube
I did this PR only so Haris and his team has direct access to the kubernetes config files.

bobeal:
Can you add a reference in the main README (explaining it is still experimental)?

kubernetes/README.md
```diff
@@ -0,0 +1,23 @@
+Tested on a Minikube cluster (for the moment).
```
bobeal, line 1: at the end, it would deserve a short note saying where (IP / port) you can contact the services

PR #122, opened 2020-07-17
fix(search): store datasetId among temporalEntityAttribute params #101

HoucemKacem:


search-service/src/main/kotlin/com/egm/stellio/search/service/TemporalEntityAttributeService.kt
```diff
@@ -56,6 +56,7 @@ class TemporalEntityAttributeService(
             .bind(
                 "entity_payload",
                 temporalEntityAttribute.entityPayload?.let { Json.of(temporalEntityAttribute.entityPayload) })
+            .bind("dataset_id", temporalEntityAttribute.datasetId)
```
vraybaud, line 59: as it is an Uri shouldn't it be converted to a string ?

HoucemKacem: The cast isn't required.

PR #124, opened 2020-07-20
feat(entity): add support for multi attribute delete attributes #101

HoucemKacem:


entity-service/src/main/kotlin/com/egm/stellio/entity/repository/Neo4jRepository.kt
```diff
@@ -255,48 +275,88 @@ class Neo4jRepository(
...
             OPTIONAL MATCH (rel)-[:HAS_VALUE]->(propOfRel:Property)
             WITH rel, propOfRel
             OPTIONAL MATCH (rel)-[:HAS_OBJECT]->(relOfRel:Relationship)
             DETACH DELETE rel, propOfRel, relOfRel
-        """.trimIndent()
+            """.trimIndent()
+        else if (datasetId == null)
+            """
+            MATCH (entity:${subjectNodeInfo.label} { id: ${'$'}entityId })-[:HAS_OBJECT]->(rel:Relationship)-[:$relationshipType]->()
+            WHERE NOT EXISTS (rel.datasetId)
```
vraybaud, line 342: deleteAll is this query without this specific line, couldn't we use that to reduce redundancy ?

entity-service/src/main/kotlin/com/egm/stellio/entity/repository/Neo4jRepository.kt
```diff
@@ -255,48 +275,88 @@ class Neo4jRepository(
...
          * Delete :
          *
          * 1. the property
          * 2. the properties of the property
          * 3. the relationships of the property
          */
-        val query = """
-            MATCH (entity:Entity { id: ${'$'}entityId })-[:HAS_VALUE]->(prop:Property { name: ${'$'}propertyName })
+
+        val query = if (deleteAll)
```
bobeal, line 287: your Cypher queries (in this function and in the next one) only differ by the initial match, you could avoid some duplication

entity-service/src/main/kotlin/com/egm/stellio/entity/repository/Neo4jRepository.kt
```diff
@@ -255,48 +275,88 @@ class Neo4jRepository(
...
+            """
+            MATCH (entity:${subjectNodeInfo.label} { id: ${'$'}entityId })-[:HAS_VALUE]->(prop:Property { name: ${'$'}propertyName })
+            OPTIONAL MATCH (prop)-[:HAS_VALUE]->(propOfProp:Property)
```
vraybaud, line 290: no way, there must be a way to avoid duplicating the below part everywhere, in both property & relationships method, maybe another method could fit..

```
OPTIONAL MATCH (prop)-[:HAS_VALUE]->(propOfProp:Property)
...
DETACH DELETE prop, propOfProp, relOfProp
```

entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityService.kt
```diff
@@ -717,21 +717,30 @@ class EntityService(
...
+                return neo4jRepository.deleteEntityProperty(
+                    subjectNodeInfo = EntitySubjectNode(entityId), propertyName = expandedAttributeName, deleteAll = deleteAll) >= 1
+            else if (neo4jRepository.hasRelationshipOfType(
+                    EntitySubjectNode(entityId), expandedAttributeName.toRelationshipTypeName()))
+                return neo4jRepository.deleteEntityRelationship(subjectNodeInfo = EntitySubjectNode(entityId),
+                    relationshipType = expandedAttributeName.toRelationshipTypeName(), deleteAll = deleteAll) >= 1
+            throw ResourceNotFoundException("Attribute $attributeName not found in entity $entityId")
+        } else {
+            if (neo4jRepository.hasPropertyInstance(EntitySubjectNode(entityId), expandedAttributeName, datasetId))
+                return neo4jRepository.deleteEntityProperty(EntitySubjectNode(entityId), expandedAttributeName, datasetId, deleteAll) >= 1
```
vraybaud, line 734: if deleteAll is by default false and you know it's false, could it be better to remove it from the method call?

entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityService.kt
```diff
@@ -717,21 +717,30 @@ class EntityService(
...
-            EntitySubjectNode(entityId),
-            expandedAttributeName.toRelationshipTypeName())
-        )
-            return neo4jRepository.deleteEntityRelationship(
-                entityId,
-                expandedAttributeName.toRelationshipTypeName()
-            ) >= 1
-
-        throw ResourceNotFoundException("Attribute $attributeName not found in entity $entityId")
+        if (deleteAll) {
```
vraybaud, line 723: i'm not a big fan of flags for handling that kind of behaviour... maybe decoupling in methods and regroup equivalent code in private method could help for future changes

bobeal: `deleteAll` will still be a flag ... not sure of what you will gain here?

entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityService.kt
```diff
@@ -717,21 +717,30 @@ class EntityService(
...
+            if (neo4jRepository.hasPropertyInstance(EntitySubjectNode(entityId), expandedAttributeName, datasetId))
+                return neo4jRepository.deleteEntityProperty(EntitySubjectNode(entityId), expandedAttributeName, datasetId, deleteAll) >= 1
+            else if (neo4jRepository.hasRelationshipInstance(
+                    EntitySubjectNode(entityId), expandedAttributeName.toRelationshipTypeName(), datasetId))
+                return neo4jRepository.deleteEntityRelationship(
+                    EntitySubjectNode(entityId), expandedAttributeName.toRelationshipTypeName(), datasetId, deleteAll) >= 1
+            datasetId?.let { throw
+            ResourceNotFoundException("Instance with datasetId $datasetId of attribute $attributeName not found in entity $entityId") }
+
+            throw ResourceNotFoundException("Default instance of $attributeName not found in entity $entityId")
```
bobeal, line 742: Quite a strange construct. A basic `if (datasetId != null) throw ... else throw ...` would be more readable.

entity-service/src/test/kotlin/com/egm/stellio/entity/repository/Neo4jRepositoryTests.kt
```diff
@@ -464,31 +464,111 @@ class Neo4jRepositoryTests {
...
+            mutableListOf(
+                Property(name = "firstName", value = "Scalpa"),
+                Property(name = "firstName", value = "Scalpa2", datasetId = URI.create("urn:ngsi-ld:Dataset:firstName:2")),
+                Property(name = "lastName", value = "Charity")
+            )
+        )
+
+        neo4jRepository.deleteEntityProperty(EntitySubjectNode(entity.id), "firstName")
+
+        assertNull(entityRepository.findById(entity.id).get().properties.find {
```
bobeal, line 497: Arf.

bobeal: Also check it has not deleted the one with a datasetId

entity-service/src/test/kotlin/com/egm/stellio/entity/repository/Neo4jRepositoryTests.kt
```diff
@@ -464,31 +464,111 @@ class Neo4jRepositoryTests {
...
+            mutableListOf(
+                Property(name = "firstName", value = "Scalpa"),
+                Property(name = "firstName", value = "Scalpa2", datasetId = URI.create("urn:ngsi-ld:Dataset:firstName:2")),
+                Property(name = "lastName", value = "Charity")
+            )
+        )
+
+        neo4jRepository.deleteEntityProperty(EntitySubjectNode(entity.id), "firstName", URI.create("urn:ngsi-ld:Dataset:firstName:2"))
+
+        assertNull(entityRepository.findById(entity.id).get().properties.find {
```
bobeal, line 519: Same as above (for both comments)

entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityService.kt
vraybaud: I don't know the "real" gain, but in my mind is clearer to have 2 methods ```deleteEntityAttribute``` and ```deleteAllEntityAttributes```.. it removes the ```if-else``` part which reduces a bit the complexity.

vraybaud: And deleteAll could have a reduced number of parameters as datasetId is not taken into account.

entity-service/src/main/kotlin/com/egm/stellio/entity/repository/Neo4jRepository.kt
HoucemKacem: We create a deleteEntityAttribute method, we pass another argument to know if the attribute is a property or a relationship ?

entity-service/src/main/kotlin/com/egm/stellio/entity/repository/Neo4jRepository.kt
HoucemKacem: A method that returns this:
```
OPTIONAL MATCH (prop)-[:HAS_VALUE]->(propOfProp:Property)
...
DETACH DELETE prop, propOfProp, relOfProp
```
to be concatenated to the initial match ?

entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityService.kt
HoucemKacem: indeed

entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityService.kt
bobeal: OK, I see what you meant. Looks good.

entity-service/src/test/kotlin/com/egm/stellio/entity/repository/Neo4jRepositoryTests.kt
HoucemKacem: > Arf.

Sorry i didn't understand what you mean

entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityService.kt
```diff
@@ -718,23 +718,37 @@ class EntityService(
...
+                relationshipType = expandedAttributeName.toRelationshipTypeName(), deleteAll = true) >= 1
 
         throw ResourceNotFoundException("Attribute $attributeName not found in entity $entityId")
     }
 
+    @Transactional
+    fun deleteEntityAttribute(entityId: String, attributeName: String, datasetId: URI?, contextLink: String): Boolean {
+        val expandedAttributeName = expandJsonLdKey(attributeName, contextLink)!!
+
+        if (neo4jRepository.hasPropertyInstance(EntitySubjectNode(entityId), expandedAttributeName, datasetId))
```
vraybaud, line 739: This makes me wonder about one thing, couldn't we do a neo4jRepository.deleteEntityAttribute which tries to delete a relation OR a property and you simply check the result to see if it was destroyed .. ? this removes the checks before trying to delete.

entity-service/src/test/kotlin/com/egm/stellio/entity/repository/Neo4jRepositoryTests.kt
bobeal: > > Arf.
> 
> Sorry i didn't understand what you mean

It is a very complex expression, hard to read and follow.

entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityService.kt
HoucemKacem: In the deleteEntityAttribute you will try first to delete it as a property, check the result and if no node was deleted try as Relationship ?
I don't think we will gain a lot here (since in some cases will end up with the same number of queries).

entity-service/src/test/kotlin/com/egm/stellio/entity/repository/Neo4jRepositoryTests.kt
HoucemKacem: Any ideas about a simpler expression ?

entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityService.kt
vraybaud: No I meant to do it in a single query. Try to delete the attribute as a Property AND a Relationships

entity-service/src/test/kotlin/com/egm/stellio/entity/repository/Neo4jRepositoryTests.kt
bobeal: At least, split it. For instance, create an intermediate variable containing the properties, and then do your check on the properties. As a bonus, you'll have one less `findById(...)`.

entity-service/src/test/kotlin/com/egm/stellio/entity/service/EntityServiceTests.kt
```diff
@@ -1305,6 +1303,80 @@ class EntityServiceTests {
...
+
+    @Test
+    fun `it should not delete all entity attribute instances if the attribute is not found`() {
+
+        val entityId = "urn:ngsi-ld:Beehive:123456"
+
+        every { neo4jRepository.hasPropertyOfName(any(), any()) } returns false
+        every { neo4jRepository.hasRelationshipOfType(any(), any()) } returns false
+
+        assertThrows<ResourceNotFoundException>("Attribute fishNumber not found in entity urn:ngsi-ld:Beehive:123456") {
```
bobeal, line 1360: Beware that the message displayed here is not checked against the message raised by the exception, it is just an informative message in case the assert fails.

If you want to check the message raised by the method, you have to do something like:

```
val exception = assertThrows...
assertEquals("my expected message", exception.message)
```

shared/src/main/kotlin/com/egm/stellio/shared/model/ExpandedEntity.kt
```diff
@@ -145,4 +145,4 @@ class ExpandedEntity private constructor(
         }
         return null
     }
-}
+}
```
bobeal, line 148: remove the noise

shared/src/main/kotlin/com/egm/stellio/shared/util/NgsiLdParsingUtils.kt
```diff
@@ -490,4 +490,4 @@ fun String.isTime(): Boolean =
         true
     } catch (e: Exception) {
         false
-    }
+    }
```
bobeal, line 493: remove the noise

entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityService.kt
```diff
@@ -718,23 +718,37 @@ class EntityService(
         neo4jRepository.deleteEntity(entityId)
 
     @Transactional
-    fun deleteEntityAttribute(entityId: String, attributeName: String, contextLink: String): Boolean {
+    fun deleteEntityAttributes(entityId: String, attributeName: String, contextLink: String): Boolean {
```
bobeal, line 721: This function name (`deleteEntityAttributes`) and name of the following (`deleteEntityAttribute`) do not clearly expose the intent of the function.

For instance, `deleteEntityAttributes` delete one attribute, so the plural form of attribute is misleading.

I would suggest:
- deleteEntityAttribute for this function (it deletes an attribute and all of its instances if it has some)
- deleteEntityAttributeInstance for the following one (it only deletes an instance of an attribute)

entity-service/src/test/kotlin/com/egm/stellio/entity/repository/Neo4jRepositoryTests.kt
```diff
@@ -464,32 +464,129 @@ class Neo4jRepositoryTests {
...
+
+    @Test
+    fun `it should delete all relationship instances`() {
         val sensor = createEntity("urn:ngsi-ld:Sensor:1233", listOf("Sensor"), mutableListOf())
         val device = createEntity("urn:ngsi-ld:Device:1233", listOf("Device"), mutableListOf())
         createRelationship(EntitySubjectNode(sensor.id), EGM_OBSERVED_BY, device.id)
+        createRelationship(EntitySubjectNode(sensor.id), EGM_OBSERVED_BY, device.id, URI.create("urn:ngsi-ld:Dataset:observedBy:01"))
 
-        neo4jRepository.deleteEntityRelationship(sensor.id, "OBSERVED_BY")
+        neo4jRepository.deleteEntityRelationship(EntitySubjectNode(sensor.id), "OBSERVED_BY", null, true)
```
bobeal, line 544: use `toRelationshipTypeName()` to avoid hardcoding such strings

entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityService.kt
HoucemKacem: The problem here is that we cannot do smth like that: 
```
OPTIONAL MATCH (propEntity:${subjectNodeInfo.label} { id: ${'$'}entityId })-[:HAS_VALUE]->(attribute:Property { name: ${'$'}propertyName })
OPTIONAL MATCH (relEntity:${subjectNodeInfo.label} { id: ${'$'}entityId })-[:HAS_OBJECT]->(attribute:Relationship)-[:${attributeName.toRelationshipTypeName()}]->()
```
```
OPTIONAL MATCH (attribute)-[:HAS_VALUE]->(propOfAttribute:Property)
WITH attribute, propOfAttribute
OPTIONAL MATCH (attribute)-[:HAS_OBJECT]->(relOfAttribute:Relationship)
DETACH DELETE attribute, propOfAttribute, relOfAttribute
```
We need to declare two variables property in the first match and relationship in the second one, which results the following cypher query:
```
OPTIONAL MATCH (propEntity:${subjectNodeInfo.label} { id: ${'$'}entityId })-[:HAS_VALUE]->(property:Property { name: ${'$'}propertyName })
OPTIONAL MATCH (relEntity:${subjectNodeInfo.label} { id: ${'$'}entityId })-[:HAS_OBJECT]->(relationship:Relationship)-[:${attributeName.toRelationshipTypeName()}]->()
```
Then in the delete query
```
OPTIONAL MATCH (property)-[:HAS_VALUE]->(propOfProperty:Property)
WITH property, propOfProperty
OPTIONAL MATCH (property)-[:HAS_OBJECT]->(relOfProperty:Relationship)

OPTIONAL MATCH (relationship)-[:HAS_VALUE]->(propOfRelationship:Property)
WITH relationship, propOfRelationship
OPTIONAL MATCH (relationship)-[:HAS_OBJECT]->(relOfRelationship:Relationship)

DETACH DELETE property, propOfProperty, relOfProperty, relationship, propOfRelationship, relOfRelationship
```
It returns property variable not found if we want to delete a relationship.

Issue #134, opened 2020-07-27
Search service fails to return the temporalEntityAttributeId with a null datasetId


HoucemKacem:
The function: getForEntityAndAttribute in TemporalEntityAttributeService

The error: datasetId is not a valid identifier. Should be of the pattern '\$([\d]+)'

Reason: we are binding the dataset_id event if it's null (so it's not added to the select statement).

PR #135, opened 2020-07-27
fix(search): fix behaviour of getForEntityAndAttribute with null datasetId #134

HoucemKacem:


search-service/src/main/kotlin/com/egm/stellio/search/service/TemporalEntityAttributeService.kt
```diff
@@ -164,13 +164,13 @@ class TemporalEntityAttributeService(
...
+        val query = databaseClient
             .execute(selectQuery)
             .bind("entity_id", id)
             .bind("attribute_name", attributeName)
-            .bind("dataset_id", datasetId)
-            .map(rowToId)
-            .one()
+
+        if (datasetId != null) return query.bind("dataset_id", datasetId).map(rowToId).one()
+        return query.map(rowToId).one()
```
bobeal, line 173: FWIW:

```
        return databaseClient
            .execute(selectQuery)
            .bind("entity_id", id)
            .bind("attribute_name", attributeName)
            .let {
                if (datasetId != null)
                    it.bind("dataset_id", datasetId)
                else it
            }
            .map(rowToId)
            .one()
```

PR #150, opened 2020-08-24
feat(common): improve ktlint integration and integrate detekt 

bobeal:


entity-service/config/detekt/baseline.xml
```diff
@@ -0,0 +1,40 @@
+<?xml version="1.0" ?>
```
HoucemKacem, line 1: i didn't get it, it's used for what ?

entity-service/src/main/kotlin/com/egm/stellio/entity/EntityServiceApplication.kt
```diff
@@ -8,6 +8,7 @@ import org.springframework.scheduling.annotation.EnableAsync
 @EnableAsync
 class EntityServiceApplication
 
+@Suppress("SpreadOperator")
```
HoucemKacem, line 11: And this tag

README.md
```diff
@@ -110,13 +110,24 @@ If you want to build only one of the services, you can launch:
...
+* You can generate the corresponding ktLint settings with the following command:
+
+```shell script
+./gradlew ktlintApplyToIdea
+```
+
+* You may consider using a plugin like [Save Actions](https://plugins.jetbrains.com/plugin/7642-save-actions) 
+that applies changed code refactoring and optimized imports on a save.
+
+* You can enable Detekt support with the [Detekt plugin](https://github.com/detekt/detekt-intellij-plugin).
```
HoucemKacem, line 130: Once enabled, how can we perform a check ?

entity-service/config/detekt/baseline.xml
bobeal: it is explained here: https://detekt.github.io/detekt/baseline.html

(we'll fix these existing code smells later, but at least, we won't add new ones)

entity-service/src/main/kotlin/com/egm/stellio/entity/EntityServiceApplication.kt
bobeal: it disables the spread operator warning for this method (https://detekt.github.io/detekt/performance.html#spreadoperator), which is not really relevant here.

README.md
bobeal: With the plugin enabled, checks are automatically done and warnings are displayed in the IDE.

entity-service/config/detekt/baseline.xml
HoucemKacem: So having code smells does not break the build

HoucemKacem (approved):

entity-service/config/detekt/baseline.xml
bobeal: existing code smells won't break it (but we'll have to fix them), new ones will break it.

PR #152, opened 2020-08-24
fix(subscription): fix warning message when a subscription query contains an unknown attribute #144

HoucemKacem:
The  `"node.get(it) must not be null"` warning is returned when a matching subscription has an unknown attribute in its query.

### DIFF
--- a/subscription-service/src/main/kotlin/com/egm/stellio/subscription/service/SubscriptionService.kt
+++ b/subscription-service/src/main/kotlin/com/egm/stellio/subscription/service/SubscriptionService.kt
@@ -420,7 +420,6 @@ class SubscriptionService(
             try {
                 runQuery(query, entity)
             } catch (e: Exception) {
-                logger.warn(e.message)
                 false
             }
         }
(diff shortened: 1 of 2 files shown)


bobeal:
> The `"node.get(it) must not be null"` warning is returned when a matching subscription has an unknown attribute in its query.

I don't really understand it. Can you explain it with an example?

Also, that does not seem to be real warning? I don't get why we have hundreds of such messages in the logs...

HoucemKacem:
> I don't really understand it. Can you explain it with an example?

For example, we create a subscription to a beeHive having a query: `unknownTemperatureAttribute>40`.

When we update the temperature of a beeHive, the unknownTemperatureAttribute (retrieved from the query) does not match any attribute in the entity payload.

And we raise an exception [Here](https://github.com/stellio-hub/stellio-context-broker/blob/develop/subscription-service/src/main/kotlin/com/egm/stellio/subscription/utils/QueryUtils.kt#L143)

HoucemKacem:
> Also, that does not seem to be real warning? I don't get why we have hundreds of such messages in the logs...

Indeed, I think better to just return false (unmatching query).

bobeal:
> > Also, that does not seem to be real warning? I don't get why we have hundreds of such messages in the logs...
> 
> Indeed, I think better to just return false (unmatching query).

Yes, that's not a problem of us if an user created a subscription that does not match.

bobeal:
OK for this version. For later, think about an improved version without throwing exceptions.

bobeal (approved):

PR #159, opened 2020-08-26
Add common NGSI-LD test suite.

jason-fox:
I have added Stellio Broker to the [FIWARE NGSI-LD_TestSuite](https://travis-ci.org/github/FIWARE/NGSI-LD_TestSuite/builds/721288348) - you can compare the results with Scorpio and Orion-LD.

This PR adds a set of common NGSI-LD compliance tests for your own use.

- Loads common compliance tests
- retrieves latest docker images
- runs tests  - current result [here](https://travis-ci.org/github/jason-fox/stellio-context-broker/builds/721294686)

You'll need to sign up for travis to get it to run obviously.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,22 @@
+language: node_js
+node_js:
+  - 12.14
+
+services:
+  - docker
+
+install : >
+  curl -LO https://github.com/FIWARE/NGSI-LD_TestSuite/archive/master.zip;
+  shasum -a 256 master.zip;
+  unzip -q master.zip;
+  rm master.zip;
+  cd NGSI-LD_TestSuite-master;  
+  export EXCLUDED_TESTS="003|004|";
+  npm install;
+  docker-compose --log-level ERROR -f docker-compose/stellio.yml pull --ignore-pull-failures 
+
+script:  >
+   echo -e " \033[1;33mRunning NGSI-LD Integration Test Suite";
+   ./services stellio;
+   export TEST_ENDPOINT=http://localhost:8080;
+   npm test;


jason-fox:
This is effectively an [ACID Test](https://en.wikipedia.org/wiki/Acid3) for context brokers. Obviously until each broker achieves full compliance the test suite is designed to fail. To pass as green with partial compliance (and remain useful to you), you will need to exclude failing tests using the `EXCLUDED_TESTS` env which is a pipe `|` separated list. 

```bash
export EXCLUDED_TESTS="003|004|";
```

Each test is numbered some are not necessary 1.3.1 compliant (e.g. `109`). To challenge the expectation of a test just raise an issue [here](https://github.com/FIWARE/NGSI-LD_TestSuite/issues)

jason-fox:
Ideally the line:

```bash
docker-compose --log-level ERROR -f docker-compose/stellio.yml pull --ignore-pull-failures 
```

should be expanded/replaced to use locally created Docker images.

bobeal:
Hi Jason,

First of all, thanks a lot for adding Stellio in the list of supported brokers for the FIWARE NGSI-LD Test Suite.

I'm keeping this PR open while we are analyzing the results, fixing the quick ones, and defining the list of the "expected" fails.

Btw, do you know if there is a way to print the HTTP requests and responses of each test in the suite?

jason-fox:
Just look for `.toHaveProperty('statusCode',` in the test suite: https://github.com/FIWARE/NGSI-LD_TestSuite/search?q=.toHaveProperty%28%27statusCode%27%2C&unscoped_q=.toHaveProperty%28%27statusCode%27%2C

Easiest way to get more output would be to amend `http.js` [here](https://github.com/FIWARE/NGSI-LD_TestSuite/blob/master/http.js#L27) and add a `console` statement.

PR #160, opened 2020-08-26
Make Docker Build standalone

jason-fox:
This PR makes the docker build reproducible by other users without the need to run Gradle prior to running docker. The code is obtained directly before building and the final build is **Distroless**

- Add standard metadata build `LABEL` (see [opencontainers.org](https://github.com/opencontainers/opencontainers.org))
- Create multi-stage build process  using a **Distroless** build:

     -    Remove [disto](https://medium.com/@dwdraju/distroless-is-for-security-if-not-for-size-6eac789f695f)
     -    Run as anonymous user

There is not much size difference compared with Alpine, but there is a security side-effect. You can no longer run bash within the container

```console
docker exec -it stellio /bin/bash
```
```text
OCI runtime exec failed: exec failed: container_linux.go:349: starting container process caused "exec: \"/bin/bash\": stat /bin/bash: no such file or directory": unknown
```

### DIFF
--- a/api-gateway/Dockerfile
+++ b/api-gateway/Dockerfile
@@ -1,7 +1,85 @@
-FROM adoptopenjdk/openjdk11:alpine-jre
-RUN addgroup -S stellio && adduser -S stellio -G stellio
-USER stellio:stellio
-ARG JAR_FILE=build/libs/*.jar
-COPY ${JAR_FILE} app.jar
-ENTRYPOINT ["java","-jar","/app.jar"]
+ARG GITHUB_ACCOUNT=stellio-hub
+ARG GITHUB_REPOSITORY=stellio-context-broker
+ARG DOWNLOAD=latest
+ARG SOURCE_BRANCH=develop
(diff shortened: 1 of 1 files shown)


jason-fox:
This is effectively just moving the `gradle` build statement inside Docker, but it makes the `Dockerfile` easy to understand. You can also use Docker `build-args` to build from alternate GitHub repos and branches.

bobeal:
Hi Jason,

We integrated your suggestions in https://github.com/stellio-hub/stellio-context-broker/pull/171.

Indeed, we are using [Jib](https://github.com/GoogleContainerTools/jib) to build Stellio's Docker images. It is a project supported by Google that allows to build optimized Docker and OCI images (see announcement [here](https://cloudplatform.googleblog.com/2018/07/introducing-jib-build-java-docker-images-better.html)). It also allows easy integration in the CI workflow as it has a Gradle plugin.

So we currently do not plan to maintain production ready Dockerfiles in Stellio repository, the ones given are just for test purposes.

Issue #157, opened 2020-08-26
Unable to read @context as a single value:


jason-fox:
The following NGSI-LD request should be valid (note that `@context` is a string not an array)

```console
curl -L -X POST 'http://stellio:8080/ngsi-ld/v1/entities/' \
-H 'Content-Type: application/ld+json' \
--data-raw '{
    "id": "urn:ngsi-ld:Building:farm001",
    "type": "Building",
    "category": {
    	"type": "Property",
        "value": ["farm"]
    },
    "address": {
        "type": "Property",
        "value": {
            "streetAddress": "Großer Stern 1",
            "addressRegion": "Berlin",
            "addressLocality": "Tiergarten",
            "postalCode": "10557"
        },
        "verified": {
			"type": "Property",
			"value": true
		}
    },
    "location": {
        "type": "GeoProperty",
        "value": {
             "type": "Point",
             "coordinates": [13.3505, 52.5144]
        }
    },
    "name": {
        "type": "Property",
        "value": "Victory Farm"
    },
    "@context": "http://context-provider:3000/data-models/ngsi-context.jsonld"
}'
```

Fails with:

```json
{
    "detail": "Cannot deserialize instance of `java.util.ArrayList<java.lang.Object>` out of VALUE_STRING token\n at [Source: (String)\"\"http://context-provider:3000/data-models/ngsi-context.jsonld\"\"; line: 1, column: 1]",
    "type": "https://uri.etsi.org/ngsi-ld/errors/BadRequestData",
    "title": "The request includes input data which does not meet the requirements of the operation"
}
```

A single string context is valid according to the [JSON-LD specification](https://w3c.github.io/json-ld-syntax/#example-6-loading-a-relative-context)

PR #163, opened 2020-08-27
fix: improve reliability when reading @context in a JSON-LD input #157

bobeal:


shared/src/main/kotlin/com/egm/stellio/shared/util/JsonLdUtils.kt
```diff
@@ -117,18 +117,7 @@ object JsonLdUtils {
         }
 
         // TODO find a way to avoid this extra parsing
```
HoucemKacem, line 119: I think the comment is deprecated now.

bobeal: no. the code is extracted in a separate method, but the extra parsing is still a thing.

HoucemKacem (approved):

PR #166, opened 2020-09-03
feat(entity): add support for (OAuth2) client authorization #167

bobeal:


entity-service/src/test/resources/ngsild/authorization/ClientCreateEvent.json
```diff
@@ -0,0 +1,10 @@
+{
+  "operationType": "CREATE",
+  "entityId": "urn:ngsi-ld:Client:191a6f0d-df07-4697-afde-da9d8a91d954",
+  "entityType": "Client",
+  "payload": "{\"id\":\"urn:ngsi-ld:Client:191a6f0d-df07-4697-afde-da9d8a91d954\",\"type\":\"Client\",\"clientId\":{\"type\":\"Property\",\"value\":\"stellio-client\"},\"@context\":[\"https://raw.githubusercontent.com/easy-global-market/ngsild-api-data-models/master/authorization/jsonld-contexts/authorization.jsonld\",\"http://uri.etsi.org/ngsi-ld/v1/ngsi-ld-core-context.jsonld\"]}",
```
HoucemKacem, line 5: I don't get the clientId property

HoucemKacem (approved):

entity-service/src/test/resources/ngsild/authorization/ClientCreateEvent.json
bobeal: It is the client_id, the one we use when configuring a client. Not used as part of the authorization process, but handy to have in the graph if you have to search for a client. Similar to the username of an User actually.

entity-service/src/main/kotlin/com/egm/stellio/entity/authorization/Neo4jAuthorizationRepository.kt
```diff
@@ -51,7 +54,13 @@ class Neo4jAuthorizationRepository(
             MATCH (userEntity:Entity { id: ${'$'}userId })
             OPTIONAL MATCH (userEntity)-[:HAS_VALUE]->(p:Property { name:"$EGM_ROLES" })
             OPTIONAL MATCH (userEntity)-[:HAS_OBJECT]-(r:Attribute:Relationship)-
-                [:IS_MEMBER_OF]->(group:Entity)-[:HAS_VALUE]->(pgroup:Property {name: "$EGM_ROLES"})
+                [:IS_MEMBER_OF]->(group:Entity)-[:HAS_VALUE]->(pgroup:Property { name: "$EGM_ROLES" })
+            RETURN p, pgroup
+            UNION
+            MATCH (client:Client)-[:HAS_VALUE]->(sid:Property { name: "$SERVICE_ACCOUNT_ID", value: ${'$'}userId })
```
HoucemKacem, line 60: Since the value of the serviceAccountId property is stored under this format: urn:ngsi-ld:Client:serviceAccountId
I think that calling this method from the service with adding the USER_PREFIX to the SUB will not match any Client

PR #169, opened 2020-09-03
Fix/end of line

philippestoltz:


.gitattributes
```diff
@@ -0,0 +1,30 @@
+# Java sources
```
bobeal, line 1: can't we say .bat is CRLF, everything else if LF?

philippestoltz: Yes, I saw that you can have a default behaviour and use overrides like this : 
* text=auto
*.bat text eol=crlf
I 'll use that

bobeal (approved):

HoucemKacem (approved):

philippestoltz:
@bobeal Sorry , I forgot to squash before merging

PR #184, opened 2020-09-10
fix(entity): update entity properties in a single cypher query

HoucemKacem:


bobeal (approved):
OK for this version. Let's keep thinking of better ways to handle such cases :)

philippestoltz (approved):

PR #189, opened 2020-09-17
feat(search): store entity location in attribute instances upon entit…

### LINKED ISSUE #2: Extract and store location of each attribute instance
Upon the reception of an entity event from Kafka, extract the location of the entity and add it to each created attribute instance.

philippestoltz:
feat(search): store entity location in attribute instances upon entity event reception closes #2

### DIFF
--- a/search-service/src/main/kotlin/com/egm/stellio/search/model/AttributeInstance.kt
+++ b/search-service/src/main/kotlin/com/egm/stellio/search/model/AttributeInstance.kt
@@ -10,5 +10,7 @@ data class AttributeInstance(
     val instanceId: URI = "urn:ngsi-ld:Instance:${UUID.randomUUID()}".toUri(),
     val observedAt: ZonedDateTime,
     val value: String? = null,
-    val measuredValue: Double? = null
+    val measuredValue: Double? = null,
+    val latitude: Double? = null,
+    val longitude: Double? = null
 )
--- a/search-service/src/main/kotlin/com/egm/stellio/search/service/EntityEventListenerService.kt
+++ b/search-service/src/main/kotlin/com/egm/stellio/search/service/EntityEventListenerService.kt
@@ -121,6 +121,7 @@ class EntityEventListenerService(
             return
         }
         val jsonLdUpdatedEntity = addContextToPayload(updatedEntity, contexts)
+        val ngsiLdEntity = expandJsonLdEntity(jsonLdUpdatedEntity).toNgsiLdEntity()
         val rawAttributeValue = attributeValuesNode["value"]
         val parsedAttributeValue =
             if (rawAttributeValue.isNumber)
@@ -132,11 +133,14 @@ class EntityEventListenerService(
         temporalEntityAttributeService.getForEntityAndAttribute(
             entityId, expandedAttributeName, datasetId
         ).zipWhen {
+            val (longitude, latitude) = ngsiLdEntity.extractCoordinatesFromLocationPoint()
             val attributeInstance = AttributeInstance(
                 temporalEntityAttribute = it,
                 observedAt = ZonedDateTime.parse(attributeValuesNode["observedAt"].asText()),
                 value = parsedAttributeValue.first,
-                measuredValue = parsedAttributeValue.second
+                measuredValue = parsedAttributeValue.second,
+                longitude = longitude,
+                latitude = latitude
             )
             attributeInstanceService.create(attributeInstance)
                 .then(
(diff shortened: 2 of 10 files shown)


search-service/src/main/kotlin/com/egm/stellio/search/listener/EntityListener.kt
```diff
@@ -28,63 +27,69 @@ class EntityListener(
...
-                        Pair(null, valueToDoubleOrNull(rawAttributeValue.asDouble()))
-                    else
-                        Pair(valueToStringOrNull(rawAttributeValue.asText()), null)
-                val datasetId = attributeValuesNode["datasetId"]?.asText()
+    private fun processUpdateEvent(entityEvent: EntityEvent) {
+        val rawEntity = expandJsonLdEntity(entityEvent.updatedEntity!!)
+        val jsonLdAttributes = JsonLdUtils.expandJsonLdFragment(entityEvent.payload!!, rawEntity.contexts)
+        val ngsiLdProperty = parseToNgsiLdAttributes(jsonLdAttributes)[0] as NgsiLdProperty
+        val propertyName = ngsiLdProperty.name
+        val temporalPropertyInstanceFound = ngsiLdProperty.instances.find { it.isTemporalAttribute() }
```
HoucemKacem, line 53: Maybe better to call it temporalPropertyInstances since you are filtering temporal instances

search-service/src/main/kotlin/com/egm/stellio/search/listener/EntityListener.kt
```diff
@@ -28,63 +27,69 @@ class EntityListener(
...
+        var temporalPropertyInstance: NgsiLdPropertyInstance?
```
HoucemKacem, line 54: Why it is declared here ? and not in the if block ? why var and not val ?

search-service/src/main/kotlin/com/egm/stellio/search/listener/EntityListener.kt
```diff
@@ -28,63 +27,69 @@ class EntityListener(
...
+        if (temporalPropertyInstanceFound.nonEmpty()) {
+            temporalPropertyInstance = temporalPropertyInstanceFound.orNull()
+        } else {
+            logger.info("Ignoring update event for $propertyName, it has no observedAt information")
+            return
+        }
+
+        val (value, measuredValue) =
```
HoucemKacem, line 62: Maybe this should go to the if block as well

search-service/src/main/kotlin/com/egm/stellio/search/listener/EntityListener.kt
philippestoltz: i 'll use just one "val" variable temporalPropertyInstance since there will be just one instance

search-service/src/main/kotlin/com/egm/stellio/search/listener/EntityListener.kt
philippestoltz: i 'll use just one "val" variable temporalPropertyInstance since there will be just one instance

search-service/src/main/kotlin/com/egm/stellio/search/listener/EntityListener.kt
philippestoltz: since we use return when temporalPropertyInstance is not found , there 's no need to put it in a if or else branch

HoucemKacem (approved):

bobeal:
We need to talk about this feature.

Mainly, adding location information in the temporal queries results is not correct. For instance, it breaks the expected temporal values representation.

philippestoltz:
> We need to talk about this feature.
> 
> Mainly, adding location information in the temporal queries results is not correct. For instance, it breaks the expected temporal values representation.

As discussed, previously I'll remove the reading of the latitude and longitude to keep the NGSI-LD compliant with the specs

search-service/src/main/kotlin/com/egm/stellio/search/listener/EntityListener.kt
```diff
@@ -28,63 +27,66 @@ class EntityListener(
...
-                val parsedAttributeValue =
-                    if (rawAttributeValue.isNumber)
-                        Pair(null, valueToDoubleOrNull(rawAttributeValue.asDouble()))
-                    else
-                        Pair(valueToStringOrNull(rawAttributeValue.asText()), null)
-                val datasetId = attributeValuesNode["datasetId"]?.asText()
+    private fun processUpdateEvent(entityEvent: EntityEvent) {
+        val rawEntity = expandJsonLdEntity(entityEvent.updatedEntity!!)
+        val jsonLdAttributes = JsonLdUtils.expandJsonLdFragment(entityEvent.payload!!, rawEntity.contexts)
+        val ngsiLdProperty = parseToNgsiLdAttributes(jsonLdAttributes)[0] as NgsiLdProperty
```
philippestoltz, line 51: @bobeal instead of retrieve location from the payload, shall I retrieve it from the entityEvent.updatedEntity ?

Issue #51, opened 2020-05-30
Upgrade to Spring Boot 2.3


bobeal:
Starting a list of things to deal with when upgrading:
- https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#howto-testcontainers

PR #203, opened 2020-09-26
feat(common): upgrade to Spring Boot 2.3 #51

bobeal:


search-service/src/main/resources/application.properties
```diff
@@ -19,13 +19,17 @@ spring.security.oauth2.resourceserver.jwt.jwk-set-uri = https://sso.eglobalmark.
 
 spring.cloud.stream.kafka.binder.brokers = localhost:29092
 spring.kafka.bootstrap-servers = localhost:29092
+# To ensure we get all past messages when dynamically joining a new topic based on our "cim.entities.*" pattern
+# By default, new matching topics are checked every 5 minutes but it can be configured
```
philippestoltz, line 23: Could you explain ? is it a new spring kafka feature ?

philippestoltz (approved):

search-service/src/main/resources/application.properties
bobeal: Explain the explanation? 🤔 What is not clear in it?

And, no, it's not new, we already have the same setting in subscription service. I spotted it during the configuration review following the upgrade.

HoucemKacem (approved):

search-service/src/main/resources/application.properties
philippestoltz: Sorry I don't know very much Kafka, how old the past messages could be ?

PR #210, opened 2020-09-29
feat(search): add support for lastN parameter in temporal entity retrieving #209

bobeal:


search-service/src/main/kotlin/com/egm/stellio/search/web/TemporalEntityHandler.kt
```diff
@@ -193,5 +193,17 @@ internal fun buildTemporalQuery(params: MultiValueMap<String, String>): Temporal
         else
             null
 
-    return TemporalQuery(params["attrs"].orEmpty(), timerel, time, endTime, params.getFirst("timeBucket"), aggregate)
+    val lastN = params.getFirst("lastN")?.toIntOrNull()?.let {
+        if (it >= 1) it else null
```
philippestoltz, line 197: I would prefer a  it > 0 test instead of it >=1  or 
if (it == 0) null else it

philippestoltz (approved):

HoucemKacem (approved):

Issue #188, opened 2020-09-15
Review the usages of InvalidRequest and BadRequestData error responses


bobeal:
It seems like there has been some mismatch between these 2 error responses and some BadRequestData errors are sent where it is expected to send InvalidRequest errors (and maybe the opposite is also true).

HoucemKacem:
- InvalidRequest that should be a BadRequestData:
There is no InvalidRequest that should be changed to a BadRequestData since we are not returning InvalidRequest exceptions.

- BadRequestData that should be a InvalidRequest:

    1. Currently, if we encounter a JsonParseException (from com.fasterxml.jackson.core) we return a JsonParseErrorResponse 
```
data class JsonParseErrorResponse(override val detail: String) : ErrorResponse(
    ErrorType.BAD_REQUEST_DATA.type,
    "The request includes invalid input data, an error occurred during JSON parsing",
    detail
)
```
From 5.5.4, an InvalidRequest should be returned, then i suggest the following changes:
```
data class JsonParseErrorResponse(override val detail: String) : ErrorResponse(
    ErrorType.INVALID_REQUEST.type,
    "The request includes invalid input data, an error occurred during JSON parsing",
    detail
)
```

bobeal:
LGTM

PR #220, opened 2020-10-12
fix(shared): throw an InvalidRequestException when request payload is not a valid JSON document #188

HoucemKacem:


shared/src/main/kotlin/com/egm/stellio/shared/util/JsonLdUtils.kt
```diff
@@ -102,6 +104,8 @@ object JsonLdUtils {
                 throw BadRequestDataException("Unable to parse input payload")
             expandedList[0]
         } catch (e: Exception) {
+            if (e is JsonParseException)
```
HoucemKacem, line 107: I added this check since the try block throws JsonParse exceptions and JsonLdParse exceptions.
- For JsonParseException, we throw the InvalidRequestException
- Else, we throw BadRequestDataException

bobeal:
I've not looked at all the code details, but one thing worries me: the main code changed are in `JsonLdUtils` but the tests are made at the handler level (via MockkedHandler and some fake "json" routes).

You should first test the new functions in JsonLdUtils.

You can add tests at the handler level, for instance by mocking JsonLdUtils and throwing JsonXXXExceptions. Then you can test you get the correct error response. But I don't see the point of adding those fake routes?

HoucemKacem:
I added the fake routes in shared module since the fix is a cross concern.
I can either:
- Add tests at the handler level of entity-service, by mocking service functions to throw the expected exceptions, but this should be done in other services as well since they are concerned by the fix.
- Keep the tests at the mocked handler of shared module, throw exceptions at the handler level without using JsonLdUtils functions.                 

I'll add tests for the new JsonLdUtils functions

shared/src/test/kotlin/com/egm/stellio/shared/util/JsonLdUtilsTests.kt
```diff
@@ -95,4 +98,42 @@ class JsonLdUtilsTests {
...
+            {
+                "id": "urn:ngsi-ld:Device:01234",,
+                "type": "Device"
+            }
+            """.trimIndent()
+
+        val exception = assertThrows<InvalidRequestException> {
+            parseAndExpandJsonLdFragment(rawEntity)
+        }
+        Assertions.assertNotNull(exception.message)
```
bobeal, line 115: why aren't you testing the content of the message (as in the next test)?

shared/src/test/kotlin/com/egm/stellio/shared/util/JsonLdUtilsTests.kt
```diff
@@ -95,4 +98,42 @@ class JsonLdUtilsTests {
...
+            """.trimIndent()
+
+        val exception = assertThrows<BadRequestDataException> {
+            parseAndExpandJsonLdFragment(rawEntity)
+        }
+        Assertions.assertEquals(
+            "Unexpected error while parsing payload : loading remote context failed: unknownContext",
+            exception.message
+        )
+    }
```
bobeal, line 138: no way to also the third error case in `parseAndExpandJsonLdFragment`?

shared/src/test/kotlin/com/egm/stellio/shared/util/MockkedHandler.kt
```diff
@@ -3,14 +3,32 @@ package com.egm.stellio.shared.util
...
 
 @RestController
 @RequestMapping("/router/mockkedroute")
-class MockkedHandler {
+class MockkedHandler : customExceptionHandler() {
 
     @PostMapping(consumes = [MediaType.APPLICATION_JSON_VALUE, JSON_LD_CONTENT_TYPE])
     fun post() = ResponseEntity.status(HttpStatus.CREATED).build<String>()
+
+    @PostMapping("/validate-json-ld")
```
bobeal, line 17: you don't really need both endpoints, no?

shared/src/test/kotlin/com/egm/stellio/shared/util/JsonLdUtilsTests.kt
HoucemKacem: Added

shared/src/test/kotlin/com/egm/stellio/shared/util/JsonLdUtilsTests.kt
HoucemKacem: I added a test for the third case

shared/src/test/kotlin/com/egm/stellio/shared/util/MockkedHandler.kt
HoucemKacem: Indeed, I updated the MockedHandlerTests [here](https://github.com/stellio-hub/stellio-context-broker/pull/220/commits/71d454d905a7d48469e1cfa1f98f6d3871f75b73)

bobeal (approved):

philippestoltz (approved):

PR #222, opened 2020-10-16
feat(entity): migrate web handlers to Kotlin coroutines #221

bobeal:


philippestoltz (approved):

entity-service/src/main/kotlin/com/egm/stellio/entity/web/APIResponses.kt
```diff
@@ -16,5 +16,6 @@ data class BatchOperationResult(
 data class BatchEntityError(
     val entityId: URI,
     val error: MutableList<String>
-
 )
+
+const val ENTITY_NOT_FOUND_MESSAGE = "Entity Not Found"
```
HoucemKacem, line 21: I think it can be moved to shared (in existing ApiUtils or new ApiResponses under shared/web) where we can add also other responses (i.e SUBSCRIPTION_NOT_FOUND_MESSAGE)

bobeal: Indeed, done, because it is also used by search service (but subscription specific messages will stay in subscription service).

HoucemKacem (approved):

Issue #238, opened 2020-10-29
Get Temporal evolution : only temporal attributes of the entity must be returned by default


philippestoltz:

philippestoltz:
without attrs it does not seem to work

PR #242, opened 2020-11-02
fix(search): only temporal attributes must be retrieved #238

philippestoltz:


search-service/src/main/kotlin/com/egm/stellio/search/web/TemporalEntityHandler.kt
```diff
@@ -160,20 +163,16 @@ internal fun buildTemporalQuery(params: MultiValueMap<String, String>): Temporal
     if (timerelParam == "between" && endTimeParam == null)
         throw BadRequestDataException("'endTime' request parameter is mandatory if 'timerel' is 'between'")
 
-    val (timerel, time) = if (timerelParam == null && timeParam == null) {
```
bobeal, line 163: Not sure the code is easier to read and maintain in this new version...

philippestoltz: it should be better now

shared/src/main/kotlin/com/egm/stellio/shared/util/ApiUtils.kt
```diff
@@ -15,11 +18,11 @@ object ApiUtils {
     }
 }
 
-fun String.parseTimeParameter(errorMsg: String): ZonedDateTime =
+fun String.parseTimeParameterwithEither(errorMsg: String): Either<String, ZonedDateTime> =
```
HoucemKacem, line 21: Maybe you can keep the old function name parseTimeParameter

HoucemKacem (approved):

shared/src/main/kotlin/com/egm/stellio/shared/util/ApiUtils.kt
philippestoltz: OK

bobeal (approved):

PR #247, opened 2020-11-09
fix(search): return empty arrays for temporal attributes when values are out of scope

philippestoltz:
#237 
Note that this fix is not ideal, TemporalEntityAttributeService.injectTemporalValues() should be refactored to avoid post processing entity temporal attributes that have no match with the injected values


search-service/src/main/kotlin/com/egm/stellio/search/service/TemporalEntityAttributeService.kt
```diff
@@ -307,6 +307,31 @@ class TemporalEntityAttributeService(
...
 
+        emptyTemporalAttributesArrays(entity, withTemporalValues)
         return JsonLdEntity(entity, jsonLdEntity.contexts)
     }
+
+    private fun emptyTemporalAttributesArrays(entity: MutableMap<String, Any>, withTemporalValues: Boolean) {
+        // temporal attributes must have empty arrays if their injected temporal evolutions are empty
+        entity.filter {
+            it.value is List<*> &&
+                (it.value as List<*>).size == 1
```
HoucemKacem, line 318: I don't really get what is happening here.
Is this the check to know if a temporal evolution is empty ?

search-service/src/main/kotlin/com/egm/stellio/search/service/TemporalEntityAttributeService.kt
```diff
@@ -307,6 +307,31 @@ class TemporalEntityAttributeService(
             entity[it.key] = it.value
         }
 
+        emptyTemporalAttributesArrays(entity, withTemporalValues)
         return JsonLdEntity(entity, jsonLdEntity.contexts)
     }
+
+    private fun emptyTemporalAttributesArrays(entity: MutableMap<String, Any>, withTemporalValues: Boolean) {
```
HoucemKacem, line 314: I think we should find a better name for this function.

search-service/src/main/kotlin/com/egm/stellio/search/web/TemporalEntityHandler.kt
```diff
@@ -120,9 +120,22 @@ class TemporalEntityHandler(
             jsonLdEntityWithTemporalValues.contexts
         )
 
-        return ResponseEntity.status(HttpStatus.OK).body(serializeObject(filteredJsonLdEntity.compact()))
+        // WORKAROUND:
+        // the json-ld lib shows empty attributes as empty objects but we need empty arrays as stated in 5.7.4.4
+        // replace empty Maps by empty Lists
+        val postProcessedEntity = replaceEmptyMapsByEmptyLists(filteredJsonLdEntity.compact())
```
HoucemKacem, line 126: Maybe you can avoid declaring this variable and put the comments above the replaceEmptyMapsByEmptyLists function.

search-service/src/main/kotlin/com/egm/stellio/search/web/TemporalEntityHandler.kt
```diff
@@ -120,9 +120,22 @@ class TemporalEntityHandler(
...
+
+        return ResponseEntity.status(HttpStatus.OK).body(serializeObject(postProcessedEntity))
     }
 
+    private fun replaceEmptyMapsByEmptyLists(compactedEntity: Map<String, Any>): Map<String, Any> {
```
HoucemKacem, line 131: it could be moved to jsonLdUtils.

search-service/src/main/kotlin/com/egm/stellio/search/service/TemporalEntityAttributeService.kt
philippestoltz: What about emptyTemporalAttritbutesHavingValueField ? (empty is a verb here)

search-service/src/main/kotlin/com/egm/stellio/search/service/TemporalEntityAttributeService.kt
philippestoltz: I am checking if there is a temporal attribute left having the property NGSILD_PROPERTY_VALUE that should have been removed. Such temporal attribute should have NGSILD_PROPERTY_VALUES instead

search-service/src/main/kotlin/com/egm/stellio/search/web/TemporalEntityHandler.kt
philippestoltz: If I move this function to jsonLdUtils the comment about the workaround should stay where it's called

search-service/src/test/resources/ngsild/beehive_with_empty_temporal_attribute_evolution.jsonld
```diff
@@ -0,0 +1,17 @@
+{
+  "id" : "urn:ngsi-ld:BeeHive:TESTC",
+  "type" : "BeeHive",
+  "incoming" : { },
```
bobeal, line 4: if you apply the same logic than for temporalValues, I would expect `incoming: []` instead

search-service/src/main/kotlin/com/egm/stellio/search/service/TemporalEntityAttributeService.kt
```diff
@@ -307,6 +307,31 @@ class TemporalEntityAttributeService(
             entity[it.key] = it.value
         }
 
+        // temporal attributes that have no injected values mpty arrays if their injected temporal evolutions are empty
```
bobeal, line 310: this comment has a problem :)

shared/src/main/kotlin/com/egm/stellio/shared/util/JsonLdUtils.kt
```diff
@@ -329,6 +329,15 @@ fun CompactedJsonLdEntity.toKeyValues(): Map<String, Any> {
     return this.mapValues { (_, value) -> simplifyRepresentation(value) }
 }
 
+fun CompactedJsonLdEntity.replaceEmptyMapsByLists(): Map<String, Any> {
```
bobeal, line 332: this is very specific to search service, so it could be left in the search service to keep this one for generic / common purposes

search-service/src/main/kotlin/com/egm/stellio/search/web/TemporalEntityHandler.kt
```diff
@@ -120,7 +120,12 @@ class TemporalEntityHandler(
             jsonLdEntityWithTemporalValues.contexts
         )
 
-        return ResponseEntity.status(HttpStatus.OK).body(serializeObject(filteredJsonLdEntity.compact()))
+        // WORKAROUND:
+        // the json-ld lib shows empty attributes as empty objects but we need empty arrays as stated in 5.7.4.4
+        // replace empty Maps by empty Lists
+        val postProcessedEntity = filteredJsonLdEntity.compact().replaceEmptyMapsByLists()
```
bobeal, line 126: I don't really understand why there is a first process in TemporalEntityAttributeService and another one here.

PR #250, opened 2020-11-13
feat(search): handle attributeUpdateEvents in search service #249

HoucemKacem:


search-service/src/test/kotlin/com/egm/stellio/search/service/EntityEventListenerServiceTest.kt
```diff
@@ -197,8 +165,113 @@ class EntityEventListenerServiceTest {
...
+        val content = prepareAttributeUpdateEventPayload(eventPayload)
+
+        entityEventListenerService.processMessage(content)
+
+        verify {
+            temporalEntityAttributeService.getForEntityAndAttribute(any(), any()) wasNot Called
+        }
+    }
+
+    private fun assertAttributeInstanceCreation(content: String) {
```
bobeal, line 239: this function name is a bit misleading because it not only asserts but also prepares the mocks and makes the call to the function under test. I would keep the concerns in separate functions (the code of the test functions will also be easier to read then).

bobeal:
Also, one last point, in the entity event listener tests, a review of the function names to clearly identify which ones are dealing with update events and which ones are dealing with replace events would be nice.

bobeal (approved):

search-service/src/test/kotlin/com/egm/stellio/search/service/EntityEventListenerServiceTest.kt
bobeal: I've made some updates in the test class to illustrate my previous comment.

philippestoltz (approved):

PR #252, opened 2020-11-19
feat(search): handle attributeAppend events in search service #251

HoucemKacem:


search-service/src/main/kotlin/com/egm/stellio/search/service/EntityEventListenerService.kt
```diff
@@ -127,4 +142,51 @@ class EntityEventListenerService(
...
+        temporalEntityAttributeService.create(temporalEntityAttribute).zipWhen {
+            attributeInstanceService.create(attributeInstance).then(
+                temporalEntityAttributeService.updateEntityPayload(entityId, updatedEntity)
+            )
+        }
+            .doOnError {
+                logger.error("Failed to persist new temporal entity attribute, ignoring it", it)
+            }.doOnNext {
+                logger.debug("Created new temporal entity attribute with one attribute instance")
+            }.subscribe()
```
bobeal, line 190: We have a very similar business logic in `TemporalEntityAttributeService`, no easy way to mutualize code?

search-service/src/test/kotlin/com/egm/stellio/search/service/EntityEventListenerServiceTest.kt
```diff
@@ -280,6 +386,18 @@ class EntityEventListenerServiceTest {
         }
     }
 
+    private fun prepareAttributeAppendEventPayload(payload: String): String =
```
bobeal, line 389: if you make `operationType` a function parameter, you should be able to have to just one `prepareXXXEventPayload` function instead of 3?

search-service/src/main/kotlin/com/egm/stellio/search/service/EntityEventListenerService.kt
HoucemKacem: Just checked, the attributeValueType is not created the same way.
Also, in createEntityTemporalReferences we are not doing exactly the same thing since we are creating the EntityPayload but here we are calling the updateEntityPayload function.

search-service/src/test/kotlin/com/egm/stellio/search/service/EntityEventListenerServiceTest.kt
HoucemKacem: Indeed

search-service/src/main/kotlin/com/egm/stellio/search/service/EntityEventListenerService.kt
bobeal: OK, let's keep it for a refactoring session later

bobeal (approved):

PR #272, opened 2020-12-14
fix: add log for NotificationService

philippestoltz:


subscription-service/src/main/kotlin/com/egm/stellio/subscription/service/NotificationService.kt
```diff
@@ -54,19 +54,21 @@ class NotificationService(
             subscriptionId = subscription.id,
             data = buildNotifData(entity, subscription.notification)
         )
-
-        if (subscription.notification.endpoint.uri.toString() == "embedded-firebase") {
+        val uri = subscription.notification.endpoint.uri.toString()
+        logger.info("Notification is about to be sent to $uri")
```
HoucemKacem, line 58: I'm wondering if you could log the success case rather than this one.

philippestoltz: Yes, you're right, I'll add a log for success

subscription-service/src/main/kotlin/com/egm/stellio/subscription/service/NotificationService.kt
```diff
@@ -54,21 +54,25 @@ class NotificationService(
...
                 request = request.header(it.key, it.value)
             }
             return request
                 .bodyValue(notification)
                 .exchange()
+                .doOnError { e -> logger.error("Failed to send notification to $uri : ${e.message}") }
                 .map {
-                    Triple(subscription, notification, it.statusCode() == HttpStatus.OK)
+                    val success = it.statusCode() == HttpStatus.OK
+                    logger.info("The notification sent has been received with ${if (success) "success" else "failure"}")
```
HoucemKacem, line 74: Is it supposed to be logged as an info when it fails ?

philippestoltz: The error would be the response of the subscriber, not an error from Stellio. The error on the execution request is already logged in doOnError method.

HoucemKacem (approved):

bobeal (approved):

Issue #286, opened 2020-12-31
Allow to query entities by a list of identifiers


bobeal:
We currently only support passing one Entity identifier (which is obviously of limited interest...). Extend this support to allow for a list of Entity identifiers (as described in 5.7.2).

PR #288, opened 2021-01-05
feat: query entities by a list of identifiers #286

HoucemKacem:


entity-service/src/test/kotlin/com/egm/stellio/entity/repository/Neo4jRepositoryTests.kt
```diff
@@ -333,6 +333,25 @@ class Neo4jRepositoryTests {
         neo4jRepository.deleteEntity(entity.id)
     }
 
+    @Test
+    fun `it should return entities matching given type and ids`() {
```
bobeal, line 337: can you also write another test with a query passing a list of ids (and not only one)?

bobeal (approved):

PR #290, opened 2021-01-07
Feature: Allow CORS configuration

HoucemKacem:


entity-service/src/main/kotlin/com/egm/stellio/entity/config/WebConfig.kt
```diff
@@ -1,15 +1,26 @@
...
 
+import org.springframework.beans.factory.annotation.Value
 import org.springframework.context.annotation.Configuration
+import org.springframework.context.annotation.Profile
 import org.springframework.http.codec.ServerCodecConfigurer
+import org.springframework.web.reactive.config.CorsRegistry
 import org.springframework.web.reactive.config.EnableWebFlux
 import org.springframework.web.reactive.config.WebFluxConfigurer
 
+@Profile("!test")
```
bobeal, line 11: why this exclusion?

HoucemKacem: Because when I added allowed.origin it breaks some unit tests. `Could not resolve placeholder 'allowed.origins' in value "${allowed.origins}"`

entity-service/src/main/kotlin/com/egm/stellio/entity/config/WebConfig.kt
```diff
@@ -1,15 +1,26 @@
...
 @Configuration
 @EnableWebFlux
 class WebConfig : WebFluxConfigurer {
 
+    @Value("\${allowed.origins}")
+    val allowedOrigins: List<String>? = null
```
bobeal, line 17: this is a bit vague when read out of context.

what about `application.cors.allowed_origins`?

entity-service/src/main/kotlin/com/egm/stellio/entity/config/WebConfig.kt
HoucemKacem: I can remove the exclusion and add allowed.origins to application-test.properties

entity-service/src/main/kotlin/com/egm/stellio/entity/config/WebConfig.kt
HoucemKacem: I'll update it

entity-service/src/main/kotlin/com/egm/stellio/entity/config/WebConfig.kt
bobeal: or add a default in application.properties

HoucemKacem: done

bobeal (approved):

PR #294, opened 2021-01-11
doc: bootstrap roadmap document

bobeal:


franckLG (commented):
It only misses Aggregation functions support which is likely to happen in medium term

Issue #268, opened 2020-12-09
Implement the discovery endpoints introduced in 1.3.1 - 5.7.5 to 5.7.10


bobeal:
Discovery endpoints:

- [x] Retrieve Available Entity Types (5.7.5)
- [x] Retrieve Details of Available Entity Types (5.7.6)
- [x] Retrieve Available Entity Type Information (5.7.7)
- [ ] Retrieve Available Attributes (5.7.8)
- [ ] Retrieve Details of Available Attributes (5.7.9)
- [ ] Retrieve Available Attribute Information (5.7.10)

bobeal:
@HoucemKacem  5.7.6 is also done, isn't it?

HoucemKacem:
> @HoucemKacem 5.7.6 is also done, isn't it?

Yes, it is done, I updated the checklist.

PR #297, opened 2021-01-13
feat(entity): implement retrieve available entity type information endpoint #268

HoucemKacem:


shared/src/main/kotlin/com/egm/stellio/shared/util/JsonLdUtils.kt
```diff
@@ -410,6 +410,14 @@ fun String.extractShortTypeFromExpanded(): String =
      */
     this.substringAfterLast("/").substringAfterLast("#")
 
+fun String.isExpanded(): Boolean =
```
HoucemKacem, line 413: Not that sure about that.
Is there a more efficient way to check if a String is a Fully Qualified Name ?

entity-service/src/main/kotlin/com/egm/stellio/entity/model/EntityTypeInfo.kt
```diff
@@ -0,0 +1,18 @@
...
+
+data class EntityTypeInfo(
+    val id: URI,
+    val type: String = "EntityTypeInformation",
+    val typeName: String,
+    val entityCount: Int,
+    val attributeDetails: List<AttributeInfo>
+)
+
+data class AttributeInfo(
```
bobeal, line 13: why not `Attribute` as is called in 5.2.28?

entity-service/src/main/kotlin/com/egm/stellio/entity/model/EntityTypeInfo.kt
```diff
@@ -0,0 +1,18 @@
...
+    val id: URI,
+    val type: String = "Attribute",
+    val attributeName: String,
+    val attributeTypes: List<String>
```
bobeal, line 17: this could be a list of Enums?

shared/src/main/kotlin/com/egm/stellio/shared/util/JsonLdUtils.kt
bobeal: you don't need to bother with this. just expand the type. if it's already expanded, you'll get it back as is (well, I'm almost pretty sure of this, check it to be sure).

entity-service/src/main/kotlin/com/egm/stellio/entity/model/EntityTypeInfo.kt
HoucemKacem: There is already an `Attribute` class.

entity-service/src/main/kotlin/com/egm/stellio/entity/model/EntityTypeInfo.kt
HoucemKacem: Indeed

entity-service/src/main/kotlin/com/egm/stellio/entity/model/EntityTypeInfo.kt
bobeal: so, I let you guess what you've forgotten :)

shared/src/main/kotlin/com/egm/stellio/shared/util/JsonLdUtils.kt
HoucemKacem: I confirm

entity-service/src/main/kotlin/com/egm/stellio/entity/model/EntityTypeInfo.kt
HoucemKacem: I didn't get it

bobeal:
There are 2 things I don't agree with:
- You are retrieving ALL entities of a given type to answer the query. It can be very costly if we have billions of entities of a given type. I would instead see a specific Cypher query that retrieves, collects and "deduplicates" the attributes of a node entity with a given label
- The authorization concerns do not apply in this use case. Authorizations apply to the data, not to the model (at least, not in the current implementation).

entity-service/src/main/kotlin/com/egm/stellio/entity/model/EntityTypeInfo.kt
HoucemKacem: Move it to the shared module ?

bobeal: when it's not "natural" (or hacky), say it in the code!

HoucemKacem:
> * You are retrieving ALL entities of a given type to answer the query. It can be very costly if we have billions of entities of a given type. I would instead see a specific Cypher query that retrieves, collects and "deduplicates" the attributes of a node entity with a given label

Indeed, i'll try this approach.

> * The authorization concerns do not apply in this use case. Authorizations apply to the data, not to the model (at least, not in the current implementation).

I see, i'll remove the authorization checks.

entity-service/src/main/kotlin/com/egm/stellio/entity/model/EntityTypeInfo.kt
```diff
@@ -0,0 +1,25 @@
+package com.egm.stellio.entity.model
+
+import java.net.URI
+
+data class EntityTypeInfo(
+    val id: URI,
+    val type: String = "EntityTypeInformation",
```
bobeal, line 7: not sure if it should be `EntityTypeInformation` or `EntityTypeInfo`

entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityTypeService.kt
```diff
@@ -0,0 +1,48 @@
...
+            AttributeInfo(
+                id = it.toUri(),
+                attributeName = it.extractShortTypeFromExpanded(),
+                attributeTypes = listOf(AttributeType.GeoProperty)
+            )
+        }
+
+        return EntityTypeInfo(
+            id = expandedType.toUri(),
+            typeName = expandedType.toRelationshipTypeName(),
```
bobeal, line 42: `toRelationshipTypeName()` ??

entity-service/src/main/kotlin/com/egm/stellio/entity/web/EntityTypeHandler.kt
```diff
@@ -0,0 +1,40 @@
...
+        @RequestHeader httpHeaders: HttpHeaders,
+        @PathVariable type: String
+    ): ResponseEntity<*> {
+        val contextLink = getContextFromLinkHeaderOrDefault(httpHeaders)
+        val expandedType = expandJsonLdKey(type.decode(), contextLink)!!
+
+        val entityTypeInformation = entityTypeService.getEntityTypeInformation(expandedType)
+            ?: throw ResourceNotFoundException("No entities found for type $expandedType")
+
+        return ResponseEntity.status(HttpStatus.OK).body(JsonUtils.serializeObject(entityTypeInformation))
```
bobeal, line 38: no `Link` header in the response?

entity-service/src/main/kotlin/com/egm/stellio/entity/model/EntityTypeInfo.kt
HoucemKacem: In 5.2.26 it is `EntityTypeInfo`

HoucemKacem:
> do you have the guarantee that it will ALWAYS be in this order?
I'll check this.

entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityTypeService.kt
HoucemKacem: updated with compactTerm method.

entity-service/src/main/kotlin/com/egm/stellio/entity/web/EntityTypeHandler.kt
HoucemKacem: Link header and Content-type header equals to `APPLICATION_JSON` ?

entity-service/src/main/kotlin/com/egm/stellio/entity/model/EntityTypeInfo.kt
HoucemKacem: Forget about the first comment.
I think it's `EntityTypeInformation` (also mentionned in 4.5.12)

HoucemKacem:
> > do you have the guarantee that it will ALWAYS be in this order?
> 
> I'll check this.

Can't find a solid answer but I think that they keep the same order mentionned when creating the Relationship [here](https://github.com/stellio-hub/stellio-context-broker/blob/ed1bc01bcf19d6c24d94820c8fb8666e6fce07ac/entity-service/src/main/kotlin/com/egm/stellio/entity/repository/Neo4jRepository.kt#L82).
We can retrieve all labels, and add an extra processing later to remove `"Attribute"` and `"Relationship"` labels, it's kind of hacky but we will be sure to pick the correct label.

bobeal:
> > > do you have the guarantee that it will ALWAYS be in this order?
> > 
> > 
> > I'll check this.
> 
> Can't find a solid answer but I think that they keep the same order mentionned when creating the Relationship [here](https://github.com/stellio-hub/stellio-context-broker/blob/ed1bc01bcf19d6c24d94820c8fb8666e6fce07ac/entity-service/src/main/kotlin/com/egm/stellio/entity/repository/Neo4jRepository.kt#L82).
> We can retrieve all labels, and add an extra processing later to remove `"Attribute"` and `"Relationship"` labels, it's kind of hacky but we will be sure to pick the correct label.

Still less hacky (or less magic) than hoping to always have the labels in the same order :)

entity-service/src/main/kotlin/com/egm/stellio/entity/web/EntityTypeHandler.kt
bobeal: I think so, it's a normal GET operation.

entity-service/src/main/kotlin/com/egm/stellio/entity/model/EntityTypeInfo.kt
bobeal: yes, but the data type is called EntityTypeInfo and I think that the example in the Annex uses EntityTypeInfo.

entity-service/src/main/kotlin/com/egm/stellio/entity/service/EntityTypeService.kt
```diff
@@ -0,0 +1,48 @@
...
+                attributeTypes = listOf(AttributeType.Relationship)
+            )
+        }
+        val geoPropertiesAttributeInfo = (attributesInformation["geoProperties"] as Set<String>).map {
+            AttributeInfo(
+                id = it.toUri(),
+                attributeName = compactTerm(it, contexts),
+                attributeTypes = listOf(AttributeType.GeoProperty)
+            )
+        }
```
bobeal, line 38: could not this be simplified (e.g., via a function)?

bobeal (approved):

bobeal:
Hmm, there is still the point on the DISTINCT that is not totally resolved?

HoucemKacem:
> Hmm, there is still the point on the DISTINCT that is not totally resolved?

Yes, there were a syntax error when trying to add `distinct` before returning the result so I added a filter later.

bobeal:
> I already tried with the `DISTINCT` keyword
> `RETURN entityCount, entityWithLocationCount, DISTINCT property.name as propertyName`
> But it's failing (syntax error), that's why I added the filter later.

You can look in this direction:

```
MATCH (entity:Entity:`$expandedType`) 
WITH count(entity) as entityCount 
OPTIONAL MATCH (entityWithLocation:Entity:`$expandedType`) WHERE entityWithLocation.location IS NOT NULL 
WITH entityCount, count(entityWithLocation) as entityWithLocationCount 
MATCH (entity:Entity:`$expandedType`) 
OPTIONAL MATCH (entity)-[:HAS_VALUE]->(property:Property) 
OPTIONAL MATCH (entity)-[:HAS_OBJECT]->(rel:Relationship) 
RETURN entityCount, entityWithLocationCount, collect(distinct property.name) as propertyNames, reduce(output = [], r IN collect(distinct labels(rel)) | output + r) as relationshipNames
```

For instance, on beehive type, it returns the following data:

image

entity-service/src/test/kotlin/com/egm/stellio/entity/repository/Neo4jRepositoryTests.kt
```diff
@@ -1243,8 +1244,145 @@ class Neo4jRepositoryTests {
...
+
+        assertTrue(attributesInformation.isEmpty())
+
+        neo4jRepository.deleteEntity(entity.id)
+    }
+
+    @Test
+    fun `it should retrieve entity type attributes information for two entities`() {
+        val firstEntity = createEntity(
+            "urn:ngsi-ld:Sensor:TESTC".toUri(),
```
bobeal, line 1269: I don't get why all those `Sensor`, `Sensor1`, `Sensor2`,... suddenly appeared.

HoucemKacem: There were conflicts with test data that breaks assertions.
I retried and it seems to work.
I pushed [this](https://github.com/stellio-hub/stellio-context-broker/pull/297/commits/5a0d480365660e6522b70e436666c8dd114cc1d2) fix commit

bobeal (approved):

PR #310, opened 2021-02-09
fix: publish entity events with current dataset id in update entity attributes #309

HoucemKacem:


entity-service/src/main/kotlin/com/egm/stellio/entity/web/EntityHandler.kt
```diff
@@ -284,7 +284,7 @@ class EntityHandler(
                 AttributeReplaceEvent(
                     entityId.toUri(),
                     compactTerm(updatedDetails.attributeName, contexts),
-                    extractDatasetIdFromNgsiLdAttributes(ngsiLdAttributes, updatedDetails.attributeName),
```
bobeal, line 287: is `extractDatasetIdFromNgsiLdAttributes` still used elsewhere? if not, remove it.

HoucemKacem: Indeed, removed.

bobeal (approved):

Issue #162, opened 2020-08-26
Delete incoming relationships when an entity is deleted


bobeal:
When an entity is deleted, the authorization relationships nodes (e.g. the rCanAdmin relationship) from users and groups are not deleted.

bobeal:
Subject updated because it does not concern only authorization relationships, but all incoming relationships.

Which is good news as it can be done in a generic way when an entity is deleted.

PR #342, opened 2021-03-16
fix(entity): delete incoming relationships when deleting an entity #162

bobeal:


entity-service/src/main/resources/neo4j/migrations/V06__remove_dangling_relationships.cypher
```diff
@@ -0,0 +1,3 @@
+MATCH (rel:Relationship)
```
HoucemKacem, line 1: I'm not totally sure about this migration script.
The fact that we match all relationships and we delete those not pointing to Entity, are we sure that all relationships in the graph (should) target entities ?  
Also it fails in local with ConstraintValidationFailed error.

bobeal: Hmm, yes, a relationship node always targets an entity (that's the NGSI-LD data model), are you thinking about something specific?

I gonna check for the constraint error.

bobeal: Damn, partial entities, I forgot about them...

bobeal: OK, updated

HoucemKacem: Yes I was thinking about Partial Entities.
LGTM

Issue #328, opened 2021-03-01
Send attribute update/append events with attribute instance payload as operation payload  


HoucemKacem:
Currently we send all attribute instances as operation payload.
https://github.com/stellio-hub/stellio-context-broker/blob/e4e3cacc8a5a2a03df06fe976e8cd29e69112535/entity-service/src/main/kotlin/com/egm/stellio/entity/web/EntityHandler.kt#L282-L298

PR #345, opened 2021-03-17
feat(entity): send only attribute instance payload in update / append event #328

bobeal:


HoucemKacem:
I think that the fix should also be done for append attribute events.

bobeal:
> I think that the fix should also be done for append attribute events.

Indeed, the code snippet in the issue made me think it was the only place where it was still missing. I'll fix it also then.

HoucemKacem (approved):

Issue #381, opened 2021-04-23
Allow to query entities by an id pattern


HoucemKacem:
Basic support for id pattern since RegExp shall be expressed in IEEE POSIX 1003.2 however Neo4j Cypher RegExp syntax is inherited from Java RegExp.

PR #382, opened 2021-04-23
feat(entity): add basic support for id pattern when querying entities #381

HoucemKacem:
I think that getEntities of Neo4jRepository needs a refactoring (i.e. Detekt ComplexMethod), should it be done in this PR ?


bobeal:
> I think that getEntities of Neo4jRepository needs a refactoring (i.e. Detekt ComplexMethod), should it be done in this PR ?

no, add an issue.

bobeal (approved):

Issue #409, opened 2021-05-21
Add partial support for attribute path in NGSI-LD Query Language


HoucemKacem:
To support querying on observedAt property of NGSI-LD properties.

PR #410, opened 2021-05-21
feat(entity): add support for querying entities with attribute path observedAt for properties #409

HoucemKacem:


entity-service/src/main/kotlin/com/egm/stellio/entity/util/QueryUtils.kt
```diff
@@ -161,13 +161,22 @@ object QueryUtils {
...
                 }
-                """
+                if (parsedQueryTerm.first.endsWith(".observedAt"))
+                    """
+                   EXISTS {
+                       MATCH (n)-[:HAS_VALUE]->(p:Property)
+                       WHERE p.name = '${parsedQueryTerm.first.removeSuffix(".observedAt")}'
+                       AND p.observedAt ${parsedQueryTerm.second} $comparableValue
+                   }
+                    """.trimIndent()
```
bobeal, line 171: It seems it would not be too difficult to have something more generic?

For instance, by splitting on the '.' character and use the second part on L.169.

HoucemKacem:
Opening new [PR](https://github.com/stellio-hub/stellio-context-broker/pull/414).

PR #421, opened 2021-05-26
feat(search): add partial support for query temporal entities with post #420

HoucemKacem:


subscription-service/config/detekt/baseline.xml
```diff
@@ -7,7 +7,6 @@
     <ID>MaxLineLength:SubscriptionService.kt$SubscriptionService$ AND ( string_to_array(watched_attributes, ',') &amp;&amp; string_to_array(:updatedAttributes, ',') OR watched_attributes IS NULL )</ID>
     <ID>MaxLineLength:WebSecurityTestConfig.kt$WebSecurityTestConfig : WebSecurityConfig</ID>
     <ID>ReturnCount:QueryUtils.kt$QueryUtils$fun extractGeorelParams(georel: String): Triple&lt;String, String?, String?&gt;</ID>
-    <ID>SpreadOperator:WebConfig.kt$WebConfig$(*it.toTypedArray())</ID>
```
bobeal, line 10: no file has been modified in subscription service, Detekt baseline should not have been modified

search-service/src/main/kotlin/com/egm/stellio/search/web/TemporalEntityHandler.kt
```diff
@@ -82,10 +80,10 @@ class TemporalEntityHandler(
         @RequestHeader httpHeaders: HttpHeaders,
         @RequestParam params: MultiValueMap<String, String>
     ): ResponseEntity<*> {
-        val withTemporalValues =
-            hasValueInOptionsParam(Optional.ofNullable(params.getFirst("options")), OptionsParamValue.TEMPORAL_VALUES)
         val contextLink = getContextFromLinkHeaderOrDefault(httpHeaders)
         val mediaType = getApplicableMediaType(httpHeaders)
+        val withTemporalValues =
+            hasValueInOptionsParam(Optional.ofNullable(params.getFirst("options")), OptionsParamValue.TEMPORAL_VALUES)
```
bobeal, line 86: why this line has moved? seems like noise?

search-service/src/main/kotlin/com/egm/stellio/search/web/TemporalEntityOperationsHandler.kt
```diff
@@ -0,0 +1,60 @@
...
+     */
+    @PostMapping("/query", consumes = [MediaType.APPLICATION_JSON_VALUE, JSON_LD_CONTENT_TYPE])
+    suspend fun queryEntities(
+        @RequestHeader httpHeaders: HttpHeaders,
+        @RequestBody requestBody: Mono<String>
+    ): ResponseEntity<*> {
+        val body = requestBody.awaitFirst()
+        val params = JsonUtils.deserializeObject(body)
+        val queryParams = LinkedMultiValueMap<String, String>()
+        params.forEach { queryParams.add(it.key, it.value.toString()) }
```
bobeal, line 35: since you are recreating the params map from the request body, why not just pass it to `queryUtils.queryTemporalEntities` and do the parsing / extraction in this method or at least why not having a function that does this instead of duplicating the processing in the 2 handlers?

search-service/src/main/kotlin/com/egm/stellio/search/util/QueryUtils.kt
```diff
@@ -0,0 +1,52 @@
...
+import com.egm.stellio.shared.model.CompactedJsonLdEntity
+import com.egm.stellio.shared.util.*
+import org.springframework.stereotype.Service
+import reactor.core.publisher.Flux
+import reactor.core.publisher.Mono
+import java.net.URI
+import java.util.*
+
+@Service
+class QueryUtils(
```
bobeal, line 17: It looks more like a service than what we usually put in utils folder (typically stateless functions). It even has a `@Service` annotation.

bobeal:
I pushed a commit to switch back to coroutines, let me know.

search-service/src/test/kotlin/com/egm/stellio/search/web/TemporalEntityOperationsHandlerTests.kt
```diff
@@ -0,0 +1,216 @@
...
+import org.springframework.test.context.ActiveProfiles
+import org.springframework.test.web.reactive.server.WebTestClient
+import org.springframework.web.reactive.function.BodyInserters
+import java.time.ZonedDateTime
+
+@ActiveProfiles("test")
+@WebFluxTest(TemporalEntityOperationsHandler::class)
+@Import(WebSecurityTestConfig::class)
+@WithMockUser
+class TemporalEntityOperationsHandlerTests {
```
bobeal, line 27: a lot of the tests here are globally the same than in `TemporalEntityHandlerTests`. if the code is better structured, you could here just focus on testing the specifics of this endpoint.

HoucemKacem:
> I pushed a commit to switch back to coroutines, let me know.

LGTM, I have also updated the PR.

bobeal:
No tests for the new `QueryService`?

search-service/src/main/kotlin/com/egm/stellio/search/web/TemporalEntityOperationsHandler.kt
bobeal: I think having two separate functions is cleaner : one that parses and validates the input and one that does the actual search

HoucemKacem: And the function that parses/checks request params should be called by the handler ? then, the handler uses the returned map to call queryTemporalEntities ?

HoucemKacem:
> No tests for the new `QueryService`?

Indeed, i'm adding some tests with **kotlinx-coroutines-test**.

search-service/src/main/kotlin/com/egm/stellio/search/web/TemporalEntityOperationsHandler.kt
bobeal: seems good.

bobeal:
Anything now useless that you could remove from the `TemporalXXXHandlerTests`? Now that the query params are checked in a specific function, you should have:
- all the query params validation checks in this new function
- only one test per handler checking a 400 request is returned if the validation failed

bobeal (approved):

PR #424, opened 2021-06-07
fix(entity): counting number of results with limit 0 #389

HoucemKacem:


entity-service/src/test/kotlin/com/egm/stellio/entity/repository/Neo4jSearchRepositoryTests.kt
```diff
@@ -302,6 +302,41 @@ class Neo4jSearchRepositoryTests {
...
+                "idPattern" to "^urn:ngsi-ld:Beekeeper:0.*2$",
+                "q" to ""
+            ),
+            sub,
+            page,
+            0,
+            DEFAULT_CONTEXTS
+        ).first
+
+        assertEquals(entitiesCount, 1)
```
bobeal, line 334: you could also check that it has returned only the count (and no entities)

bobeal (approved):

Issue #428, opened 2021-06-14
Replace `page` by `offset` in pagination queries


bobeal:
Wrong reading of the specification, let's use the correct `offset` request parameter!

PR #429, opened 2021-06-15
fix(common): replace page by offset in pagination queries #428

HoucemKacem:


subscription-service/src/test/kotlin/com/egm/stellio/subscription/web/SubscriptionHandlerTests.kt
```diff
@@ -305,11 +308,14 @@ class SubscriptionHandlerTests {
         every { subscriptionService.getSubscriptions(any(), any(), any()) } returns Flux.just(subscription)
 
         webClient.get()
-            .uri("/ngsi-ld/v1/subscriptions/?limit=1&page=1")
+            .uri("/ngsi-ld/v1/subscriptions/?limit=1&offset=1")
             .exchange()
             .expectStatus().isOk
             .expectHeader()
-            .valueEquals("Link", "</ngsi-ld/v1/subscriptions?limit=1&page=2>;rel=\"next\";type=\"application/ld+json\"")
+            .valueEquals(
```
bobeal, line 315: shouldn't we also have a prev with `offset=0&limit=1`?

HoucemKacem: Yes, I updated the test case to return only the next link, since we already have a test that checks both prev and next links.
I have also fixed the default offset [here](https://github.com/stellio-hub/stellio-context-broker/pull/429/commits/fff27296bc5e2a6d026103ec231e1b7a475570ff).

bobeal (approved):

Issue #517, opened 2021-11-10
add support for `timeInterval` parameter


bobeal:
see for instance Notification Behavior in 5.8.6 and Subscription data definition in table 5.2.12-1.

PR #572, opened 2022-01-26
feat(subscription):add support for time interval (#517)

PoulomiNandy:


bobeal:
do not forget to add the business checks at creation / modification time : "Greater than 0. if watchedAttributes is present, it shall not appear (0 cardinality)".

PR #663, opened 2022-05-20
refactor(common): extend queryparams data class to wrap all query parameters

gpoujol:


gpoujol:
@bobeal I applied your last suggestions

bobeal:
OK for the code review.

One last thing before merging this PR: `parseAndCheckParams` in `QueryUtils` module is missing some tests (see SonarCloud coverage report) and it is doing some tasks that can easily fail, so add a few tests for this function.

bobeal (approved):

PR #705, opened 2022-06-27
fix(entity): allow to add / remove a service account to / from a group

bobeal:


gpoujol (approved):

gpoujol:
Shouldn't the same be done for the delete? Since we add the service to the group, you need to delete the group?

bobeal:
> Shouldn't the same be done for the delete? Since we add the service to the group, you need to delete the group?

Indeed, good point.

PR #718, opened 2022-07-20
fix(search): apply pagination on entities whatever their matching attributes

gpoujol:


search-service/src/main/kotlin/com/egm/stellio/search/service/TemporalEntityAttributeService.kt
```diff
@@ -268,21 +268,27 @@ class TemporalEntityAttributeService(
     ): Mono<List<TemporalEntityAttribute>> {
         val selectQuery =
             """
+                WITH entities AS (
+                    SELECT distinct(entity_id)
```
bobeal, line 272: you have to apply the ordering when you get the list of matching entities. if you don't, you'll get unpredictable results.

search-service/src/main/kotlin/com/egm/stellio/search/service/TemporalEntityAttributeService.kt
```diff
@@ -268,21 +268,27 @@ class TemporalEntityAttributeService(
...
+                    FROM temporal_entity_attribute
+                    GROUP BY entity_id
```
bobeal, line 274: do you really need this?

search-service/src/main/kotlin/com/egm/stellio/search/service/TemporalEntityAttributeService.kt
bobeal: you also need to apply the filters in this query, you only want the list of entities that match the requested filters (type, id pattern, ...)

gpoujol: I add the ordering but I need to filters two times

gpoujol: It's necessary because when I first filter on my id entities I get a set of id entities. Then I ask for the full temporal entity attribute information with the id of the one included in my set of ids. But for example an attribute that is not wanted can have an id that is in my list so it will be returned.

search-service/src/main/kotlin/com/egm/stellio/search/service/TemporalEntityAttributeService.kt
gpoujol: Are you talking about `GROUP BY` ?

bobeal: yes

search-service/src/main/kotlin/com/egm/stellio/search/service/TemporalEntityAttributeService.kt
bobeal: hmm, ok, I see. so maybe the second filtering (on the full TEA) is lighter than the first one (only filtering on `attrs`)?

gpoujol: I wonder if it can't have the same problem for the type (multityping)

search-service/src/main/kotlin/com/egm/stellio/search/service/TemporalEntityAttributeService.kt
```diff
@@ -266,26 +266,36 @@ class TemporalEntityAttributeService(
         queryParams: QueryParams,
         accessRightFilter: () -> String?
     ): Mono<List<TemporalEntityAttribute>> {
+        val filterQuery = buildEntitiesQueryFilter(
+            queryParams,
+            accessRightFilter
+        )
+
+        val filterQueryWithPrefix = buildEntitiesQueryFilter(
```
bobeal, line 274: `filterOnAttributesQuery`? (don't get what this `WithPrefix` is)

bobeal (approved):

PR #726, opened 2022-08-01
feat(search): coroutines and API changes

bobeal:


search-service/src/main/kotlin/com/egm/stellio/search/service/IAMListener.kt
```diff
@@ -31,31 +31,53 @@ class IAMListener(
...
+                    " for entity ${authorizationEvent.entityId}"
+            )
+        })
     }
 
     @KafkaListener(topics = ["cim.iam.replay"], groupId = "search-iam-replay")
-    fun processIamReplay(content: String) {
+    suspend fun processIamReplay(content: String) {
         logger.debug("Received IAM replay event: $content")
         when (val authorizationEvent = JsonUtils.deserializeAs<EntityEvent>(content)) {
```
gpoujol, line 82: Why we don't use the same configuration than above 
`val authorizationEvent = JsonUtils.deserializeAs<EntityEvent>(content)
 when (authorizationEvent) ...`

And do not add the fold with  logger.error and logger.debug ?

gpoujol (approved):

search-service/src/main/kotlin/com/egm/stellio/search/service/IAMListener.kt
bobeal: because replay is one shot process and I did not want to take care of it...

gpoujol (approved):

PR #738, opened 2022-08-23
feat(entity): get entities with simle geo query

gpoujol:


entity-service/src/test/kotlin/com/egm/stellio/entity/repository/StandaloneNeo4jSearchRepositoryTests.kt
```diff
@@ -460,38 +518,33 @@ class StandaloneNeo4jSearchRepositoryTests : WithNeo4jContainer {
         )
         createRelationship(EntitySubjectNode(entity.id), "observedBy", partialTargetEntityUri)
 
+        val queryParams = QueryParams(
```
bobeal, line 521: this is not the same `queryParams` that was used in the calls to `getEntities` below, you are breaking things here.

entity-service/src/test/kotlin/com/egm/stellio/entity/util/GeoQueryUtilsTests.kt
```diff
@@ -0,0 +1,46 @@
...
+            georel = "near;maxDistance==1500",
+            geometry = "Point",
+            coordinates = "[57.5522, -20.3484]",
+            geoproperty = NGSILD_LOCATION_PROPERTY
+        )
+        Assertions.assertEquals(geoQuery, geoQueryParams)
+    }
+
+    @Test
+    fun `extract georel params`() {
```
bobeal, line 29: it is poorly named

shared/config/detekt/baseline.xml
```diff
@@ -7,6 +7,7 @@
     <ID>LongParameterList:NgsiLdEntity.kt$NgsiLdGeoPropertyInstance$( val coordinates: WKTCoordinates, createdAt: ZonedDateTime?, modifiedAt: ZonedDateTime?, observedAt: ZonedDateTime? = null, datasetId: URI? = null, properties: List&lt;NgsiLdProperty&gt; = emptyList(), relationships: List&lt;NgsiLdRelationship&gt; = emptyList() )</ID>
     <ID>LongParameterList:NgsiLdEntity.kt$NgsiLdPropertyInstance$( val value: Any, val unitCode: String?, createdAt: ZonedDateTime?, modifiedAt: ZonedDateTime?, observedAt: ZonedDateTime?, datasetId: URI?, properties: List&lt;NgsiLdProperty&gt;, relationships: List&lt;NgsiLdRelationship&gt; )</ID>
     <ID>LongParameterList:NgsiLdEntity.kt$NgsiLdRelationshipInstance$( val objectId: URI, createdAt: ZonedDateTime?, modifiedAt: ZonedDateTime?, observedAt: ZonedDateTime?, datasetId: URI?, properties: List&lt;NgsiLdProperty&gt;, relationships: List&lt;NgsiLdRelationship&gt; )</ID>
+    <ID>MatchingDeclarationName:GeoUtils.kt$GeoQueryUtils</ID>
```
bobeal, line 10: you should take care of this

entity-service/src/main/kotlin/com/egm/stellio/entity/repository/QueryUtils.kt
```diff
@@ -50,30 +53,53 @@ object QueryUtils {
...
+                ORDER BY id
+                SKIP ${queryParams.offset} LIMIT ${queryParams.limit}
+            """.trimIndent()
+        } else
             """
                 WITH collect(distinct(entity.id)) as entityIds, count(entity.id) as count
                 UNWIND entityIds as id
                 RETURN id, count
                 ORDER BY id
-                SKIP ${queryParams.offset} LIMIT ${queryParams.limit}
```
bobeal, line 66: I'm not sure about this

entity-service/src/main/kotlin/com/egm/stellio/entity/repository/Neo4jSearchRepository.kt
```diff
@@ -19,22 +26,47 @@ class Neo4jSearchRepository(
...
+                val distance = distance(geo1, geo2) * MULTIPLY_DISTANCE
+                if (georelParams.second.equals("<=")) {
+                    if (distance <= georelParams.third!!.toDouble()) geoResult.add(it)
+                } else if (georelParams.second.equals(">=")) {
+                    if (distance >= georelParams.third!!.toDouble()) geoResult.add(it)
+                } else if (georelParams.second.equals("==")) {
+                    if (distance(geo1, geo2) != georelParams.third!!.toDouble()) geoResult.add(it)
+                }
+            }
+            return return prepareResults(queryParams.limit, geoResult)
```
bobeal, line 67: I was suggesting to use https://neo4j.com/docs/cypher-manual/current/functions/spatial/#functions-distance, why did you choose another way?

Btw, by filtering the results here, you're breaking the pagination. 

But let's keep it like this for now...

entity-service/src/main/kotlin/com/egm/stellio/entity/repository/Neo4jSearchRepository.kt
```diff
@@ -19,22 +26,47 @@ class Neo4jSearchRepository(
...
+            geoQuery.geometry == "Point" &&
+            geoQuery.coordinates != null
+        ) {
+            val geoResult: ArrayList<Map<String, Any>> = ArrayList()
+            val geo1 = WKTReader().read(geoJsonToWkt(geoQuery.geometry, geoQuery.coordinates.toString()))
+            val georelParams = extractGeorelParams(geoQuery.georel)
+
+            result.forEach {
+                val geo2 = WKTReader().read(it["entityLocation"] as String)
+                val distance = distance(geo1, geo2) * MULTIPLY_DISTANCE
```
bobeal, line 58: in the `geoQ`, the distance is in meters. what is the unit of the result returned by `distance(...)`?

entity-service/src/main/kotlin/com/egm/stellio/entity/repository/Neo4jSearchRepository.kt
gpoujol: I found it easier to treat it like this. 
I couldn't see how to retrieve the longitude and latitude and make the query to calculate the distance at the same time.
The function requires a longitude and a latitude for its operation and I did not see how to do the transformation (wkt to geo) in the same requests

entity-service/src/main/kotlin/com/egm/stellio/entity/repository/Neo4jSearchRepository.kt
gpoujol: The result of the unit is in kilometers. That's why I did *1000

entity-service/src/main/kotlin/com/egm/stellio/entity/repository/QueryUtils.kt
gpoujol: This part was already here before

bobeal: yes, but you removed this line

gpoujol: oupss sorry

entity-service/src/main/kotlin/com/egm/stellio/entity/repository/Neo4jSearchRepository.kt
bobeal: I can't find this info in the doc of the function. where did you find it?

gpoujol: I did not find the info in their doc but by observing the results of the postgis function I get the same.
I know it's not very correct but I haven't found better solutions yet. (I wanted to keep a library that we already used)

bobeal: By "postgis function", are you referring to `ST_Distance`? (if so, it returns results in meters)

bobeal: Did a quick test, you get a distance in degrees.

entity-service/src/test/kotlin/com/egm/stellio/entity/repository/Neo4jSearchRepositoryTests.kt
```diff
@@ -104,6 +108,158 @@ class Neo4jSearchRepositoryTests : WithNeo4jContainer {
...
+            sub,
+            DEFAULT_CONTEXTS
+        ).second
+
+        assertTrue(entities.size == 1)
+        assertTrue(entities.containsAll(listOf(fourthEntity.id)))
+    }
+
+    @Test
+    fun `it should return matching entities with geo query and egalDistance`() {
```
bobeal, line 210: I'm really not sure that this "equal distance" thing exists. Where did you find it?

entity-service/src/main/kotlin/com/egm/stellio/entity/repository/QueryUtils.kt
```diff
@@ -201,4 +225,24 @@ object QueryUtils {
...
+                     point({
+                        latitude: toFloat(apoc.text.regexGroups(entity.location, '([\\d\\.]+) ([\\d\\.]+)')[0][1]), 
+                        longitude: toFloat(apoc.text.regexGroups(entity.location, '([\\d\\.]+) ([\\d\\.]+)')[0][2]),
+                        crs: 'wgs-84'
+                     }),
+                     point({
+                        latitude: ${coordinates[1].toFloat()}, 
+                        longitude: ${coordinates[2].toFloat()}, 
+                        crs: 'wgs-84'
+                     })
```
bobeal, line 244: If I refer to https://www.rfc-editor.org/rfc/rfc7946#section-3.1.1, first element is longitude, and second is latitude. Seems you are doing the opposite.

gpoujol: Yes indeed I change it

entity-service/src/test/kotlin/com/egm/stellio/entity/repository/Neo4jSearchRepositoryTests.kt
gpoujol: Not in the spec but when we do `extractGeorelParams` we put `=`. So I treat it. Maybe we put finally `Triple(georel, null, null)` instead `Triple(DISTANCE_QUERY_CLAUSE, "=", comparisonParams[1])`

bobeal: ideally, this should be rejected (as is said in the comment in the `extractGeorelParams` function). Leave as it is for now, but don't write a test for something that should not work (while inventing an `equalDistance` param)

entity-service/src/main/kotlin/com/egm/stellio/entity/repository/QueryUtils.kt
```diff
@@ -201,4 +225,24 @@ object QueryUtils {
...
+                $prefix entity.location CONTAINS '${geoQuery.geometry!!.uppercase()}'
+                AND point.distance(
+                     point({
+                        latitude: toFloat(apoc.text.regexGroups(entity.location, '([\\d\\.]+) ([\\d\\.]+)')[0][2]), 
+                        longitude: toFloat(apoc.text.regexGroups(entity.location, '([\\d\\.]+) ([\\d\\.]+)')[0][1]),
+                        crs: 'wgs-84'
+                     }),
+                     point({
+                        latitude: ${coordinates[1].toFloat()}, 
+                        longitude: ${coordinates[2].toFloat()},
```
bobeal, line 242: sure for the lat / lon order??

gpoujol: This part corresponds to the coordinates given by the user during the request. 
In this exemple request `http:///ngsi-ld/v1/entities?type=Test1&geoproperty=location&georel=near;maxDistance==1500&geometry=Point&coordinates=[57.5522, -20.3484]` the lat are 57.5522 and the lon are -20.3484 ?

bobeal: don't guess, read the spec. from 4.10 : `A coordinates (parameter name coordinates) element which shall represent the coordinates of the reference geometry as mandated by IETF RFC 7946 [8], section 3.1.1.`

PR #766, opened 2022-10-04
fix(subcription): add missing JSON-LD behabiors

gpoujol:


gpoujol:
@bobeal currently when we make the request with the q parameter this one is not expand but it works with the payload of a non-expand entity. To expand the parameter q I need to expand also the entity payload ?

bobeal:
> @bobeal currently when we make the request with the q parameter this one is not expand but it works with the payload of a non-expand entity. To expand the parameter q I need to expand also the entity payload ?

I don't understand "To expand the parameter q I need to expand also the entity payload ?".

To expand the `q` parameter, you only need a JSON-LD context, the one provided at the creation time of the subscription (which we now keep in DB IIRC).

bobeal:
Is it still a draft (if so, what is missing?) or is it reviewable?

gpoujol:
> > @bobeal currently when we make the request with the q parameter this one is not expand but it works with the payload of a non-expand entity. To expand the parameter q I need to expand also the entity payload ?
> 
> I don't understand "To expand the parameter q I need to expand also the entity payload ?".
> 
> To expand the `q` parameter, you only need a JSON-LD context, the one provided at the creation time of the subscription (which we now keep in DB IIRC).

Expand the `q` parameter is good but when we check if query matching we use compact entityPayload. I must to expand it before a

gpoujol:
> Is it still a draft (if so, what is missing?) or is it reviewable?

gpoujol:
> Is it still a draft (if so, what is missing?) or is it reviewable?

Sorry, I left it out the last days. I picked it up there. I finish some changes and it will be reviewable.

bobeal:
superseded by https://github.com/stellio-hub/stellio-context-broker/pull/818

PR #778, opened 2022-11-08
fix/search by regex query on entities

gpoujol:


gpoujol:
> The build has failed

It failed here https://github.com/stellio-hub/stellio-context-broker/blob/79620f16b1d1e7d6ea8c0e28ac00c262d734fce0/shared/src/main/kotlin/com/egm/stellio/shared/util/JsonLdUtils.kt#L116

bobeal:
> > The build has failed
> 
> It failed because of that
> 
> https://github.com/stellio-hub/stellio-context-broker/blob/79620f16b1d1e7d6ea8c0e28ac00c262d734fce0/shared/src/main/kotlin/com/egm/stellio/shared/util/JsonLdUtils.kt#L116-L118

now you have a good reason to change it.

search-service/src/main/kotlin/com/egm/stellio/search/service/TemporalEntityAttributeService.kt
```diff
@@ -428,7 +428,7 @@ class TemporalEntityAttributeService(
                         WHEN attribute_type = 'Property' AND attribute_value_type IN ('DATETIME', 'DATE', 'TIME') THEN 
                             jsonb_path_exists(temporal_entity_attribute.payload,
                                 '$."$NGSILD_PROPERTY_VALUE" ? 
-                                    (@."$JSONLD_VALUE_KW".datetime() ${query.second} $targetValue.datetime())')
+                                    (@."$JSONLD_VALUE_KW".datetime() ${query.second} $targetValue)')
```
bobeal, line 431: don't see what it has to do with the PR?

search-service/src/main/kotlin/com/egm/stellio/search/service/TemporalEntityAttributeService.kt
```diff
@@ -461,7 +461,7 @@ class TemporalEntityAttributeService(
             queryTerm.contains("<") ->
                 Triple(queryTerm.split("<")[0], "<", queryTerm.split("<")[1])
             queryTerm.contains("=~") ->
-                Triple(queryTerm.split("=~")[0], "=~", queryTerm.split("=~")[1])
+                Triple(queryTerm.split("=~")[0], "like_regex", queryTerm.split("=~")[1])
```
bobeal, line 464: did you check the compatibility of the `like_regex` operator with the regex format that should be supported by NGSI-LD implementations?

search-service/src/main/kotlin/com/egm/stellio/search/service/TemporalEntityAttributeService.kt
gpoujol: When I have this, for exemple if targetValue = "Paris" the request does this `"Paris".datetime()"' but PostgreSQL don't like that. It returns errors at the level of `.`

bobeal: This is weird since it should only go there if attribute is of type `DATETIME`, `DATE` or `TIME`. There is another problem somewhere else...

bobeal:
I was a bit annoyed with this regex and the dates, so I looked at the spec (4.9) and:

```
Match pattern (production rule named patternOp). A matching entity shall contain the target element and the
target value shall be in the L(R) of the regular pattern specified by the Query Term:
- If the target value data type is different than String then it shall be considered as not matching.
```

gpoujol:
> I was a bit annoyed with this regex and the dates, so I looked at the spec (4.9) and:
> 
> ```
> Match pattern (production rule named patternOp). A matching entity shall contain the target element and the
> target value shall be in the L(R) of the regular pattern specified by the Query Term:
> - If the target value data type is different than String then it shall be considered as not matching.
> ```

Ok I change it.

bobeal:
I just pushed a commit with some fixes:
- Swapped the two `WHEN` in `buildInnerQuery()` since the second one (for dates) was never matched as it was already matched in the first `WHEN` (a property with a date is a property)
- Moved `String.convertInDateTimeIfNeeded()` outside the class, since I noticed that, when inside the class, `this` refers to the `TemporalAttributeService` bean and not to the string on which the call is made (I guess this is something related to how Spring beans are wired, to be checked)
- Changed the way the queries on dates are made in the unit test. If passed as a string (e.g., `dateOfFirstBee=="2018-12-04T12:00:00Z"`), it is considered as a string, not as a date

The build is failing after this commit but that was "expected" after these changes.

Just saw that this expression returns false, I let you look at it:

```
jsonb_path_exists(
    temporal_entity_attribute.payload,
    '$."https://uri.etsi.org/ngsi-ld/hasValue" ? (@."@value".datetime() == "2018-12-04T12:00:00Z".datetime())'
)
```

search-service/src/test/kotlin/com/egm/stellio/search/service/TemporalEntityAttributeServiceTests.kt
```diff
@@ -315,6 +315,93 @@ class TemporalEntityAttributeServiceTests : WithTimescaleContainer, WithKafkaCon
...
+
+        assertEquals(1, temporalEntityAttributes.size)
+        assertThat(temporalEntityAttributes)
+            .allMatch {
+                it.attributeName in setOf("https://uri.etsi.org/ngsi-ld/default-context/dateOfFirstBee")
+            }
+    }
+
+    @Test
+    fun `it should no matching entities when do a regex on the parameter q with datetime value`() = runTest {
```
bobeal, line 377:
```suggestion
    fun `it should not retrieve the temporal attributes of entities when doing a regex query on a datetime`() = runTest {
```

gpoujol:
> I just pushed a commit with some fixes:
> 
> * Swapped the two `WHEN` in `buildInnerQuery()` since the second one (for dates) was never matched as it was already matched in the first `WHEN` (a property with a date is a property)
> * Moved `String.convertInDateTimeIfNeeded()` outside the class, since I noticed that, when inside the class, `this` refers to the `TemporalAttributeService` bean and not to the string on which the call is made (I guess this is something related to how Spring beans are wired, to be checked)
> * Changed the way the queries on dates are made in the unit test. If passed as a string (e.g., `dateOfFirstBee=="2018-12-04T12:00:00Z"`), it is considered as a string, not as a date
> 
> The build is failing after this commit but that was "expected" after these changes.
> 
> Just saw that this expression returns false, I let you look at it:
> 
> ```
> jsonb_path_exists(
>     temporal_entity_attribute.payload,
>     '$."https://uri.etsi.org/ngsi-ld/hasValue" ? (@."@value".datetime() == "2018-12-04T12:00:00Z".datetime())'
> )
> ```

The problems comes with `.datetime(). It returns nothing. 
I tried to use a template but I can't reuse the one I used for the `to_timestamp` like here `to_timestamp ( '2020-01-24T13:01:22.066Z', 'YYYY-MM-DD"T"HH24:MI:SS.US' )`.

bobeal:
> > I just pushed a commit with some fixes:
> > 
> > * Swapped the two `WHEN` in `buildInnerQuery()` since the second one (for dates) was never matched as it was already matched in the first `WHEN` (a property with a date is a property)
> > * Moved `String.convertInDateTimeIfNeeded()` outside the class, since I noticed that, when inside the class, `this` refers to the `TemporalAttributeService` bean and not to the string on which the call is made (I guess this is something related to how Spring beans are wired, to be checked)
> > * Changed the way the queries on dates are made in the unit test. If passed as a string (e.g., `dateOfFirstBee=="2018-12-04T12:00:00Z"`), it is considered as a string, not as a date
> > 
> > The build is failing after this commit but that was "expected" after these changes.
> > Just saw that this expression returns false, I let you look at it:
> > ```
> > jsonb_path_exists(
> >     temporal_entity_attribute.payload,
> >     '$."https://uri.etsi.org/ngsi-ld/hasValue" ? (@."@value".datetime() == "2018-12-04T12:00:00Z".datetime())'
> > )
> > ```
> 
> The problems comes with `.datetime(). It returns nothing. I tried to use a template but I can't reuse the one I used for the `to_timestamp`like here`to_timestamp ( '2020-01-24T13:01:22.066Z', 'YYYY-MM-DD"T"HH24:MI:SS.US' )`.

There are two `.datetime()`, which one returns nothing?

gpoujol:
> jsonb_path_exists(
>     temporal_entity_attribute.payload,
>     '$."https://uri.etsi.org/ngsi-ld/hasValue" ? (@."@value".datetime() == "2018-12-04T12:00:00Z".datetime())'
> )

No when I change the operator here `SELECT jsonb_path_query('["2018-12-04T12:00:00Z", "2020-01-24T13:01:22Z"]', '$[*] ? (@.datetime() > "2016-12-04T12:00:00Z".datetime())') ;`. I use `==`, `<` and `>` I have always nothing. Whereas if I remove `.datetime()` I have results. I think I can't parse the date.

shared/src/main/kotlin/com/egm/stellio/shared/util/DateUtils.kt
```diff
@@ -6,6 +6,8 @@ import java.time.format.DateTimeFormatter
 
 val formatter: DateTimeFormatter = DateTimeFormatter.ISO_INSTANT
 
+const val TEMPLATE: String = "\"YYYY-MM-DD\\\"T\\\"HH24:MI:SS.US\\\"Z\\\"\""
```
bobeal, line 9: very specific to PG, better put in in `DBUtils`. With a more meaningful name (e.g., `DATETIME_TEMPLATE`)

shared/src/testFixtures/kotlin/com/egm/stellio/shared/util/JsonLdContextUtils.kt
```diff
@@ -28,3 +28,5 @@ const val NAME_PROPERTY = "https://schema.org/name"
 const val CREATED_BY_RELATIONSHIP = "https://ontology.eglobalmark.com/egm#createdBy"
 const val MANAGED_BY_COMPACT_RELATIONSHIP = "managedBy"
 const val MANAGED_BY_RELATIONSHIP = "https://ontology.eglobalmark.com/egm#managedBy"
+const val DATE_OF_FIRST_BEE_COMPACT_PROPERTY = "dateOfFirstBee"
+const val DATE_OF_FIRST_BEE_PROPERTY = "https://uri.etsi.org/ngsi-ld/default-context/dateOfFirstBee"
```
bobeal, line 32: don't you think your URI does not look like the others? 🤔

gpoujol: I can put this instead `const val DATE_OF_FIRST_BEE_PROPERTY = "https://uri.etsi.org/ngsi-ld/default-context/$DATE_OF_FIRST_BEE_COMPACT_PROPERTY"` but otherwise I don't see

bobeal: 1. you are using the core context from NGSI-LD, this is a no-go zone for your own stuff
2. other terms are under `ontology.eglobalmark.com/...` (you'll probably have to add it in the context managed on GH, and to update the cached version in `shared/src/testFixtures/resources/contexts/apic.jsonld`)

gpoujol: How I do to update the cached ?

bobeal: hmm, the response is the point 2 in my previous commit 🤔

bobeal (approved):

PR #794, opened 2022-11-22
fix/add a default creation date for tea

gpoujol:


search-service/src/main/kotlin/db/migration/V0_28__JsonLd_migration.kt
```diff
@@ -97,7 +97,7 @@ class V0_28__JsonLd_migration : BaseJavaMigration() {
             val defaultCreatedAt = getPropertyValueFromMapAsDateTime(
                 expandedEntity as Map<String, List<Any>>,
                 NGSILD_CREATED_AT_PROPERTY
-            )!!
+            ) ?: ZonedDateTime.parse("1970-01-01T00:00:00Z")
```
bobeal, line 100: hmm, this line is already there in the `feature/data-migration-script` branch

search-service/src/main/kotlin/db/migration/V0_28__JsonLd_migration.kt
```diff
@@ -208,7 +208,7 @@ class V0_28__JsonLd_migration : BaseJavaMigration() {
         attributePayload: Map<String, List<Any>>,
         ngsiLdAttributeInstance: NgsiLdAttributeInstance
     ) {
-        val createdAt = ngsiLdAttributeInstance.createdAt
+        val createdAt = ngsiLdAttributeInstance.createdAt ?: ZonedDateTime.parse("1970-01-01T00:00:00Z")
```
bobeal, line 211: why not doing in the same way than it is done in `createTeaEntry()`?

gpoujol: I do this

bobeal (approved):

PR #798, opened 2022-11-25
fix/replace simple quote by two dollar symbol

gpoujol:


bobeal:
I don't get why you are doing the dollar-quote escaping for *every* column, even ones we know for sure they are not concerned.

gpoujol:
> I don't get why you are doing the dollar-quote escaping for _every_ column, even ones we know for sure they are not concerned.

Is not it better to apply the same thing everywhere than to make an exception?

bobeal:
> > I don't get why you are doing the dollar-quote escaping for _every_ column, even ones we know for sure they are not concerned.
> 
> Is not it better to apply the same thing everywhere than to make an exception?

Everywhere where it can happen, yes. What's the point, for instance, of doing it for an UUID or a date that has already been parsed and validated?

gpoujol:
> > > I don't get why you are doing the dollar-quote escaping for _every_ column, even ones we know for sure they are not concerned.
> > 
> > 
> > Is not it better to apply the same thing everywhere than to make an exception?
> 
> Everywhere where it can happen, yes. What's the point, for instance, of doing it for an UUID or a date that has already been parsed and validated?

Ok I will change that

search-service/src/main/kotlin/db/migration/V0_28__JsonLd_migration.kt
```diff
@@ -194,7 +194,7 @@ class V0_28__JsonLd_migration : BaseJavaMigration() {
                             ${temporalAttributesMetadata.value.measuredValue}, 
                             ${temporalAttributesMetadata.value.value.toSQLValue()}, 
                             ST_GeomFromText('${temporalAttributesMetadata.value.geoValue!!.value}'), 
-                            '$teaId', '$attributeInstanceId', '$serializedAttributePayload')
+                            '$teaId', $$$attributeInstanceId$$, '$serializedAttributePayload')
```
bobeal, line 197: `attributeInstanceId` is generated by us and does not need to be dollar-quoted. But `serializedAttributePayload` should.

search-service/src/main/kotlin/db/migration/V0_28__JsonLd_migration.kt
```diff
@@ -205,7 +205,7 @@ class V0_28__JsonLd_migration : BaseJavaMigration() {
                         ('$createdAt', '${AttributeInstance.TemporalProperty.CREATED_AT}', 
                             ${temporalAttributesMetadata.value.measuredValue}, 
                             ${temporalAttributesMetadata.value.value.toSQLValue()}, 
-                            '$teaId', '$attributeInstanceId', '$serializedAttributePayload')
+                            '$teaId', $$$attributeInstanceId$$, $$$serializedAttributePayload$$)
```
bobeal, line 208: same for `attributeInstanceId`

bobeal (approved):

PR #799, opened 2022-11-28
fix: move-function-after-attributes-exepansion

gpoujol:


bobeal:
You moved the de-duplication function after JSON-LD expansion but the tests are still done on a compacted payload. How can you be sure the de-duplication is still running fine? (I am not totally sure)

bobeal (approved):

PR #804, opened 2022-12-01
fix/problem related to context in entity payload table

gpoujol:


search-service/src/main/kotlin/db/migration/V0_28__JsonLd_migration.kt
```diff
@@ -88,7 +99,8 @@ class V0_28__JsonLd_migration : BaseJavaMigration() {
                 """
                 update entity_payload
                 set payload = $$$serializedJsonLdEntity$$,
-                    specific_access_policy = ${specificAccessPolicy.toSQLValue()}
+                    specific_access_policy = ${specificAccessPolicy.toSQLValue()},
+                    contexts = $$$contexts$$
```
bobeal, line 103: the `contexts` column is an array (of text), I don't think this gonna work. Have you ever tried it?

gpoujol: > the `contexts` column is an array (of text), I don't think this gonna work. Have you ever tried it?

I tested this morning and no it doesn't work. I tried to add `typpedArray()` but without success

gpoujol: I tried to do this 
```
jdbcTemplate.update(
                """
                update entity_payload
                set created_at = :createdAt,
                    payload = :serializedJsonLdEntity,
                    specific_access_policy = :specificAccessPolicy,
                    contexts = :contexts
                where entity_id = :entityId
                """.trimIndent(),
                mapOf(
                    "createdAt" to defaultCreatedAt,
                    "serializedJsonLdEntity" to serializedJsonLdEntity,
                    "specificAccessPolicy" to specificAccessPolicy.toSQLValue(),
                    "contexts" to contexts.toTypedArray()
                )
            )
``` 
But I have this error `Caused by: org.postgresql.util.PSQLException: L'indice de la colonne est hors limiteá: 1, nombre de colonnesá: 0.`

bobeal: The `contexts` variable is a list of strings.

At the end, you want to populate the `contexts` column with something like that: `contexts = ARRAY['context1', 'context2', ...]`.

So maybe something like `contexts = ARRAY[${contexts.joinToString(separator = ",", prefix = "'", postfix = "'")}]`

gpoujol: Thanks for the help it works much better. Just as I said in the dailly the contexts are in this form when there is only one `{https://raw.githubusercontent.com/easy-global-market/ngsild-api-data-models/master/smartCameraUseCase/jsonld-contexts/smartCamera-context.jsonld}
` and this when there are more than one  `{"https://raw.githubusercontent.com/easy-global-market/ngsild-api-data-models/master/smartCameraUseCase/jsonld-contexts/smartCamera-context.jsonld,https://uri.etsi.org/ngsi-ld/v1/ngsi-ld-core-context-v1.3.jsonld"} `. I look at this

PR #805, opened 2022-12-01
fix/add default creation date in entity_payload table

gpoujol:


search-service/src/main/kotlin/db/migration/V0_28__JsonLd_migration.kt
```diff
@@ -111,6 +97,21 @@ class V0_28__JsonLd_migration : BaseJavaMigration() {
                     ZonedDateTime.parse("1970-01-01T00:00:00Z")
                 }
 
+            // store the expanded entity payload instead of the compacted one
+            val serializedJsonLdEntity = serializeObject(expandedEntity).replace("'", "''")
+            jdbcTemplate.execute(
+                """
+                update entity_payload
+                set createdAt = $defaultCreatedAt,
```
bobeal, line 105: doesn't it need to be inside single quotes?

search-service/src/main/kotlin/db/migration/V0_28__JsonLd_migration.kt
```diff
@@ -111,6 +97,21 @@ class V0_28__JsonLd_migration : BaseJavaMigration() {
                     ZonedDateTime.parse("1970-01-01T00:00:00Z")
                 }
 
+            // store the expanded entity payload instead of the compacted one
+            val serializedJsonLdEntity = serializeObject(expandedEntity).replace("'", "''")
```
bobeal, line 101: just noticed that this `replace` may not fit well with the dollar quoting added below.

gpoujol: I can remove it with the dollar quoting it is no longer useful, no ?

PR #822, opened 2023-01-02
feat: no longer support the sys attrs request parameter

gpoujol:


bobeal:
when merging, reword the commit message (without context, one can think support for sysAttrs has been removed from  Stellio as a whole): "chore(authz): remove support for sysAttrs in get authorized entities endpoint"

bobeal (approved):

PR #823, opened 2023-01-05
feat: update bootstrap file for auth test suite

gpoujol:


search-service/src/main/resources/application-api-tests.properties
```diff
@@ -0,0 +1,6 @@
+application.apitests.userid.1=06664685-ee36-42a0-aaa7-089d8af1a864
+application.apitests.userid.2=b40f98ec-5319-416c-9f03-e4a92e153716
+application.apitests.userid.3=e5edfb90-d68d-4ec2-b779-1b171f5ef7be
+
+application.apitests.groupid.1=c8e2d969-faf6-407e-9431-02d7e4d0a01f
+application.apitests.groupid.2=82e09721-9e37-4ded-b491-158de39a875b
```
bobeal, line 6: no...

gpoujol: no need to add this file ?

bobeal: it was not there before, why should it be added now?

What's more, this is quite confidential data (ok, about tests users but anyway)

bobeal (approved):

PR #824, opened 2023-01-05
fix: error on data type output cast in SQL query

gpoujol:


search-service/src/main/kotlin/com/egm/stellio/search/authorization/SubjectReferentialService.kt
```diff
@@ -188,7 +188,9 @@ class SubjectReferentialService(
                 """.trimIndent()
             )
             .bind("subject_id", (sub as Some).value)
-            .allToMappedList { GlobalRole.forKey(it["global_roles"] as String) }
```
bobeal, line 191: instead, what about changing the SQL query with:

```
SELECT unnest(global_roles)
FROM subject_referential
WHERE subject_id = :subject_id
```

and leaving the rest of the code untouched.

also, add one (or more) unit test for this function, this kind of problem should appear before the integration tests.

gpoujol:
I added 3 tests to verify the function (for a user with no role, one role and two role).
But I wonder about the usefulness of the three . Just two should be enought ? (user with no role and with mulitple role)

search-service/src/test/kotlin/com/egm/stellio/search/authorization/SubjectReferentialServiceTests.kt
```diff
@@ -447,4 +448,55 @@ class SubjectReferentialServiceTests : WithTimescaleContainer {
                 fail("it should have returned a ResourceNotFoundException exception")
             })
     }
+
+    @Test
+    fun `it should get global roles for user with no role`() = runTest {
+        val userAccessRights = SubjectReferential(
```
bobeal, line 454: call this variable `subjectReferential`, not `userAccessRights` (and same in the two other tests)

bobeal (approved):

PR #825, opened 2023-01-05
fix : do not bind entities ids if empty in EntityAccessRightsService

gpoujol:


search-service/src/main/kotlin/com/egm/stellio/search/authorization/EntityAccessRightsService.kt
```diff
@@ -244,51 +244,55 @@ class EntityAccessRightsService(
         sub: Option<Sub>,
         entities: List<URI>
     ): Either<APIException, Map<URI, Map<AccessRight, List<SubjectRightInfo>>>> = either {
-        val subjectUuids = subjectReferentialService.getSubjectAndGroupsUUID(sub).bind()
+        if (entities.isNullOrEmpty())
+            emptyMap()
```
bobeal, line 248: you could `return` here to avoid nesting the rest of the function code into an `else` (the code is already a bit complex, it makes things harder to understand)

gpoujol: Ok

bobeal (approved):

PR #827, opened 2023-01-09
fix: expand correctly subjectInfo

gpoujol:


shared/src/main/kotlin/com/egm/stellio/shared/util/JsonLdUtils.kt
```diff
@@ -568,7 +568,7 @@ object JsonLdUtils {
      */
     fun buildNonReifiedProperty(value: Any): List<Map<String, Any>> =
         listOf(
-            mapOf(JSONLD_ID to value)
+            mapOf(JSONLD_ID to value.toString())
```
bobeal, line 571: hmm, any reason?

search-service/src/main/kotlin/com/egm/stellio/search/util/JsonLdUtils.kt
```diff
@@ -0,0 +1,59 @@
...
+     *              {
+     *                  "@value": "username"
+     *              }
+     *         ]
+     *       }
+     *     ]
+     *   }
+     * ]
+     */
+    fun buildExpandedSubjectInfo(value: Map<String, String>): ExpandedAttributePayload =
```
bobeal, line 39: this will only work is subject is an user, and not if it is a client or a group. build something more generic.

also wondering if not better to keep it into `EntityAccessRights` (very specific to this use case)

search-service/src/main/kotlin/com/egm/stellio/search/service/EntityPayloadService.kt
```diff
@@ -10,6 +10,7 @@ import com.egm.stellio.search.util.*
 import com.egm.stellio.shared.model.*
 import com.egm.stellio.shared.util.*
 import com.egm.stellio.shared.util.AuthContextModel.SpecificAccessPolicy
+import com.egm.stellio.shared.util.JsonLdUtils
```
bobeal, line 13: ?

search-service/src/test/kotlin/com/egm/stellio/search/util/QueryUtilsTests.kt
```diff
@@ -5,6 +5,8 @@ import com.egm.stellio.search.model.AttributeInstance
 import com.egm.stellio.search.model.TemporalQuery
```
bobeal, line 5: no real modifications in this file.

search-service/src/test/kotlin/com/egm/stellio/search/web/EntityAccessControlHandlerTests.kt
```diff
@@ -697,9 +711,12 @@ class EntityAccessControlHandlerTests {
             jsonLdEntity[AUTH_PROP_SAP] = JsonLdUtils.buildExpandedProperty(specificAccessPolicy)
         }
         rCanReadUser?.run {
```
bobeal, line 713: this `createJsonLdEntity` is having a lot of JSON-LD manipulation and this is not good. Maybe rewrite it to:
- create an `EntityAccessRight` object
- call its `serializeProperties` function
- returns this result wrapped in a `JsonLdEntity`

shared/src/main/kotlin/com/egm/stellio/shared/util/JsonLdUtils.kt
gpoujol: When we compact entity @id wait for a string

bobeal: I know but it should happen automatically. Did you have an error somewhere?

gpoujol: Yes, I have an error here [when we serialize datasetId in subjectInfo](https://github.com/stellio-hub/stellio-context-broker/blob/4d01f5e1542f3b4c2655a29cb214a1a1102c1376/search-service/src/main/kotlin/com/egm/stellio/search/authorization/EntityAccessRights.kt#L39)

bobeal: please, elaborate on this:
- what was the exact error?
- when exactly did it happen?

(since this function is used similarly in other places, I need to understand the problem before deciding on a change)

search-service/src/main/kotlin/com/egm/stellio/search/authorization/EntityAccessRights.kt
```diff
@@ -68,3 +78,67 @@ data class EntityAccessRights(
...
+ *              {
+ *                  "@value": "username"
+ *              }
+ *         ]
+ *       }
+ *     ]
+ *   }
+ * ]
+ */
+fun buildExpandedSubjectInfo(value: Map<String, String>): ExpandedAttributePayload =
```
bobeal, line 109: these 2 functions could be private?

search-service/src/main/kotlin/com/egm/stellio/search/authorization/EntityAccessRights.kt
```diff
@@ -68,3 +78,67 @@ data class EntityAccessRights(
...
+    listOf(
+        mapOf(
+            JSONLD_TYPE to listOf(NGSILD_PROPERTY_TYPE.uri),
+            NGSILD_PROPERTY_VALUE to listOf(
```
bobeal, line 113: why don't you just iterate over the map of subject info instead of hardcoding keys (and missing some entries btw)?

search-service/src/test/kotlin/com/egm/stellio/search/web/EntityAccessControlHandlerTests.kt
```diff
@@ -682,25 +693,33 @@ class EntityAccessControlHandlerTests {
...
+        val earSerialized = entityAccessRights.serializeProperties()
+        return JsonLdEntity(earSerialized, listOf(context))
+    }
+
+    private fun createEntityAccessRight(
+        id: URI,
+        type: ExpandedTerm,
+        accessRight: AccessRight,
+        specificAccessPolicy: AuthContextModel.SpecificAccessPolicy? = null,
+        rCanReadUser: List<EntityAccessRights.SubjectRightInfo>? = null
```
bobeal, line 708: rCanReadUsers

shared/src/main/kotlin/com/egm/stellio/shared/util/JsonLdUtils.kt
gpoujol: Ok I understand.

- The exact error is : 
       `"title": "invalid @id value",
        "detail": "invalid @id value: value of @id must be a string",
        "type": "https://uri.etsi.org/ngsi-ld/errors/BadRequestData"`

- This errors happens during compaction. It comes when we serialize `SubjectRightInfo` at the function `buildNonReifiedProperty` beacause we put a datasetid which is a uri. We associate a uri with`@id` and this is not allowed.

[It is done here](https://github.com/stellio-hub/stellio-context-broker/blob/4d01f5e1542f3b4c2655a29cb214a1a1102c1376/search-service/src/main/kotlin/com/egm/stellio/search/authorization/EntityAccessRights.kt#L39)
That's why I thought parse datasetId when we call the function rather than modify the function directly. 

We can also modify the type of datasetId at the base but I think it is less good

search-service/src/main/kotlin/com/egm/stellio/search/authorization/EntityAccessRights.kt
gpoujol: Yeah but I don't see how expand keys. 
Currently the keys are kind and username (name or subjectId) but I don't see how to put them in expanded format.

bobeal: what if you simply use `expandJsonLdFragment` to expand the whole `subjectInfo`?

shared/src/main/kotlin/com/egm/stellio/shared/util/JsonLdUtils.kt
bobeal: if the value of `@id` must be a string, then the `buildNonReifiedProperty` function should only accept a string as argument.

gpoujol: `buildExpandedRelationship` accept only an URI and parse this URI to in the function. Maybe we can do the same thing ?

bobeal (approved):

PR #828, opened 2023-01-10
feat: add username and group name to tests subjects

gpoujol:


bobeal:
Looks OK for the code but your commit message does not match what I see in the code

gpoujol:
> Looks OK for the code but your commit message does not match what I see in the code

Maybe it's better to say :`Add username and groupName for different user and group in db` ?

bobeal:
> > Looks OK for the code but your commit message does not match what I see in the code
> 
> Maybe it's better to say :`Add username and groupName for different user and group in db` ?

Add username and group name to tests subjects

bobeal (approved):

PR #831, opened 2023-01-16
fix: apply the rule 6.3.5 on authz endpoint

gpoujol:
We apply the rule 6.3.5 on authz endpoint and add authorization context in a default context. 
If the context provided by the user does not provide correct terms for post queries, these return an exception.


search-service/src/main/kotlin/com/egm/stellio/search/web/EntityAccessControlHandler.kt
```diff
@@ -245,9 +248,15 @@ class EntityAccessControlHandler(
...
 
-            val body = requestBody.awaitFirst()
-            val expandedPayload = expandJsonLdFragment(AUTH_TERM_SAP, body, COMPOUND_AUTHZ_CONTEXT)
+            val body = requestBody.awaitFirst().deserializeAsMap()
+            val contexts = checkAndGetContext(httpHeaders, body).addAuthzContextIfNeeded()
+            val rawPayload = mapOf(AUTH_TERM_SAP to JsonLdUtils.removeContextFromInput(body))
+            val expandedPayload = expandJsonLdFragment(rawPayload, contexts)
             val ngsiLdAttributes = parseToNgsiLdAttributes(expandedPayload)
+            if (ngsiLdAttributes[0].name != AUTH_PROP_SAP)
+                ResourceNotFoundException(
```
gpoujol, line 257: I'm not really sure about `ResourceNotFoundException` A `BadRequestDataException` would be more appropriate ?

gpoujol:
I wonder if there should be something for `removeRightsOnEntity` and `deleteSpecificAccessPolicy` ?

search-service/src/main/kotlin/com/egm/stellio/search/authorization/AuthorizationUtils.kt
```diff
@@ -0,0 +1,15 @@
+package com.egm.stellio.search.authorization
+
+import com.egm.stellio.shared.util.AuthContextModel.COMPOUND_AUTHZ_CONTEXT
+import com.egm.stellio.shared.util.AuthContextModel.NGSILD_AUTHORIZATION_CONTEXT
+import com.egm.stellio.shared.util.JsonLdUtils.NGSILD_CORE_CONTEXT
+
+fun List<String>.addAuthzContextIfNeeded(): List<String> =
```
bobeal, line 7: In authz endpoints, we consider the "default context" to be NGSI-LD Core Context + Authz context. Here, you are replacing the NGSI-LD Core Context by the Authz context.

search-service/src/main/kotlin/com/egm/stellio/search/web/EntityAccessControlHandler.kt
bobeal: indeed

bobeal:
> I wonder if there should be something for removeRightsOnEntity and deleteSpecificAccessPolicy ?

I remember a specific rule for delete operations where the context has no role to play (e.g., when deleting an entity by id), but unable to find it in the spec (maybe just a discussion that never landed into the spec).

But these 2 endpoints are in this category, let's just ignore the context for them.

search-service/src/main/kotlin/com/egm/stellio/search/authorization/AuthorizationUtils.kt
gpoujol: Ok for POST but when we do GET we don't have a list but a single string for context.

bobeal: yes, but you have a compound context for authz

shared/src/main/kotlin/com/egm/stellio/shared/model/NgsiLdEntity.kt
```diff
@@ -358,7 +358,7 @@ fun parseToNgsiLdAttributes(attributes: Map<String, Any>): List<NgsiLdAttribute>
     }
 
 fun String.isNgsiLdSupportedName() =
-    this.all { char -> char.isLetterOrDigit() || listOf(':', '_').contains(char) }
+    this.all { char -> char.isDefined() && !listOf('<', '"', '\'', '=', ';', '(', ')', '>', '?', '*').contains(char) }
```
bobeal, line 361: you need to rebase your branch to remove this hack commit (done temporarily for https://github.com/FIWARE/load-tests/issues/90)

shared/src/main/kotlin/com/egm/stellio/shared/util/AuthUtils.kt
```diff
@@ -21,6 +21,8 @@ val CREATION_ROLES: Set<GlobalRole> = setOf(STELLIO_CREATOR).plus(ADMIN_ROLES)
 
 object AuthContextModel {
     const val NGSILD_AUTHORIZATION_CONTEXT = "$EGM_BASE_CONTEXT_URL/authorization/jsonld-contexts/authorization.jsonld"
+    const val AUTHORIZATION_COMPOUND_CONTEXT =
+        "$EGM_BASE_CONTEXT_URL/authorization/jsonld-contexts/authorization-compound.jsonld"
     val COMPOUND_AUTHZ_CONTEXT = listOf(NGSILD_AUTHORIZATION_CONTEXT, NGSILD_CORE_CONTEXT)
```
bobeal, line 26: for clarity, a renaming would be good, since there are now 2 things called "compound something" and distinguish them is not easy.

shared/src/main/kotlin/com/egm/stellio/shared/model/NgsiLdEntity.kt
gpoujol: I don't know if the result is the same but I did a revert of this commit

search-service/src/main/kotlin/com/egm/stellio/search/authorization/AuthorizationUtils.kt
gpoujol: Maybe I could put these two functions in `AuthUtils` ?

shared/src/main/kotlin/com/egm/stellio/shared/util/AuthUtils.kt
```diff
@@ -21,7 +21,9 @@ val CREATION_ROLES: Set<GlobalRole> = setOf(STELLIO_CREATOR).plus(ADMIN_ROLES)
 
 object AuthContextModel {
     const val NGSILD_AUTHORIZATION_CONTEXT = "$EGM_BASE_CONTEXT_URL/authorization/jsonld-contexts/authorization.jsonld"
-    val COMPOUND_AUTHZ_CONTEXT = listOf(NGSILD_AUTHORIZATION_CONTEXT, NGSILD_CORE_CONTEXT)
+    const val AUTHORIZATION_COMPOUND_CONTEXT =
+        "$EGM_BASE_CONTEXT_URL/authorization/jsonld-contexts/authorization-compound.jsonld"
+    val DEFAULT_AUTHZ_CONTEXT = listOf(NGSILD_AUTHORIZATION_CONTEXT, NGSILD_CORE_CONTEXT)
```
bobeal, line 26: These 3 values are really missing consistency...

I suggest:

```
    const val AUTHORIZATION_CONTEXT = "$EGM_BASE_CONTEXT_URL/authorization/jsonld-contexts/authorization.jsonld"
    const val AUTHORIZATION_COMPOUND_CONTEXT =
        "$EGM_BASE_CONTEXT_URL/authorization/jsonld-contexts/authorization-compound.jsonld"
    val AUTHORIZATION_API_DEFAULT_CONTEXTS = listOf(...)
```

search-service/src/main/kotlin/com/egm/stellio/search/authorization/AuthorizationUtils.kt
bobeal: yes

PR #855, opened 2023-02-09
feat(common): remove costly compactions when sending entity events in Kafka

bobeal:


subscription-service/src/main/kotlin/com/egm/stellio/subscription/service/EntityEventListenerService.kt
```diff
@@ -73,16 +73,16 @@ class EntityEventListenerService(
     }
 
     private suspend fun handleEntityEvent(
-        updatedAttributes: Set<String>,
+        updatedAttributes: Set<ExpandedTerm>,
         entityPayload: String,
-        contexts: List<String>
+        contexts: List<ExpandedTerm>
```
gpoujol, line 78: Contexts is still a list of string

shared/src/main/kotlin/com/egm/stellio/shared/model/EntityEvent.kt
```diff
@@ -24,14 +24,14 @@ open class EntityEvent(
     val operationType: EventsType,
     open val sub: String?,
     open val entityId: URI,
-    open val entityTypes: List<String>,
-    open val contexts: List<String>
+    open val entityTypes: List<ExpandedTerm>,
+    open val contexts: List<ExpandedTerm>
```
gpoujol, line 28: Context is still String. Same everywhere here

gpoujol (approved):

PR #856, opened 2023-02-13
refactor: better SoC in core API services and query improvements

bobeal:
- use entity service as the main entrypoint for core API services (avoid cyclic dependencies)
- use native JSON type for data persisted as jsonb
- async delete of attributes and attributes instances (better performance)

### DIFF
--- a/search-service/src/test/kotlin/com/egm/stellio/search/util/TestUtils.kt
+++ b/search-service/src/test/kotlin/com/egm/stellio/search/util/TestUtils.kt
@@ -47,4 +47,4 @@ fun buildSapAttribute(specificAccessPolicy: AuthContextModel.SpecificAccessPolic
 }
 
 const val EMPTY_PAYLOAD = "{}"
-val EMPTY_JSON_PAYLOAD = Json.of("{}")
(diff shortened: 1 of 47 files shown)


search-service/src/test/kotlin/com/egm/stellio/search/util/TestUtils.kt
```diff
@@ -44,3 +45,6 @@ fun buildSapAttribute(specificAccessPolicy: AuthContextModel.SpecificAccessPolic
         expandJsonLdFragment(AUTH_TERM_SAP, sapPropertyFragment, AuthContextModel.AUTHORIZATION_API_DEFAULT_CONTEXTS)
     )[0]
 }
+
+const val EMPTY_PAYLOAD = "{}"
+val EMPTY_JSON_PAYLOAD = Json.of("{}")
```
gpoujol, line 50: Maybe we can write `val EMPTY_JSON_PAYLOAD = Json.of($EMPTY_PAYLOAD)`

gpoujol (approved):

bobeal:
@gpoujol only one single comment? :(

gpoujol (approved):

PR #864, opened 2023-02-20
fix: add a check on idPattern info in subscriptions

gpoujol:


subscription-service/src/main/kotlin/com/egm/stellio/subscription/service/SubscriptionService.kt
```diff
@@ -61,6 +62,8 @@ class SubscriptionService(
             checkTimeIntervalGreaterThanZero(subscription).bind()
             checkSubscriptionValidity(subscription).bind()
             checkExpiresAtInTheFuture(subscription).bind()
+            checkExpiresAtInTheFuture(subscription).bind()
```
bobeal, line 65: Duplicate check?

bobeal (approved):

PR #865, opened 2023-02-21
fix: check attribute existence by all reference or one

gpoujol:


bobeal:
- A better solution is already implemented (check the code of `deleteInstance`)
- And btw, the `deleteInstance` function is not correct (it is even easy to make it fail)

gpoujol:
> * A better solution is already implemented (check the code of `deleteInstance`)
> 
>     * And btw, the `deleteInstance` function is not correct (it is even easy to make it fail)

I don't see why it's not correct and what I can use of this function

search-service/src/main/kotlin/com/egm/stellio/search/service/TemporalEntityAttributeService.kt
```diff
@@ -654,7 +654,7 @@ class TemporalEntityAttributeService(
                     from temporal_entity_attribute 
                     where entity_id = :entity_id 
                     and attribute_name = :attribute_name
-                    ${datasetId.toDatasetIdFilter()}
+                    ${datasetId.toDatasetIdFilter()}
```
bobeal, line 657: ?

Issue #683, opened 2022-06-06
Complete support for NGSI-LD Query Language


bobeal:
Advanced features of NGSI-LD query language is not yet implemented (e.g., attribute path).

PR #869, opened 2023-02-26
feat: add support for NGSI-LD Query Language #683

bobeal:


gpoujol (approved):

search-service/src/main/kotlin/com/egm/stellio/search/util/TemporalQueryUtils.kt
```diff
@@ -1,32 +1,33 @@
...
         Pair(pagination.limitDefault, pagination.limitMax),
         requestParams,
         contextLink
-    )
+    ).bind()
+
     if (queryParams.types.isEmpty() && queryParams.attrs.isEmpty() && inQueryEntities)
-        throw BadRequestDataException("Either type or attrs need to be present in request parameters")
+        BadRequestDataException("Either type or attrs need to be present in request parameters")
+            .left().bind<TemporalEntitiesQuery>()
```
gpoujol, line 30: why add <TemporalEntitiesQuery>

search-service/src/main/kotlin/com/egm/stellio/search/util/TemporalQueryUtils.kt
```diff
@@ -36,17 +37,20 @@ fun parseAndCheckQueryParams(
...
-    val temporalQuery = buildTemporalQuery(requestParams, inQueryEntities)
+    val temporalQuery = buildTemporalQuery(requestParams, inQueryEntities).bind()
 
-    return TemporalEntitiesQuery(
+    TemporalEntitiesQuery(
         queryParams = queryParams,
         temporalQuery = temporalQuery,
         withTemporalValues = withTemporalValues,
         withAudit = withAudit
     )
```
gpoujol, line 47: there shouldn't be `.right` at the end ?

bobeal: no it is in a either effect (and actually compilation would have failed if the return type was not correct)

search-service/src/main/kotlin/com/egm/stellio/search/util/TemporalQueryUtils.kt
bobeal: I have a warning if not (but don't really know why it can't automatically infer it, we have the same thing in the web handlers)

PR #872, opened 2023-03-02
feat: implement entity type selection language for search service

gpoujol:


bobeal:
LGTM

subscription-service/src/main/kotlin/com/egm/stellio/subscription/model/EntityTypeSelector.kt
```diff
@@ -1,10 +1,9 @@
 package com.egm.stellio.subscription.model
 
-import com.egm.stellio.shared.model.ExpandedTerm
 import java.net.URI
 
-data class EntityInfo(
+data class EntityTypeSelector(
```
bobeal, line 5: I see no occurrence of `EntityTypeSelector` in the specification

subscription-service/src/main/kotlin/com/egm/stellio/subscription/service/NotificationService.kt
```diff
@@ -34,7 +34,13 @@ class NotificationService(
         either {
             val id = ngsiLdEntity.id
             val types = ngsiLdEntity.types
-            subscriptionService.getMatchingSubscriptions(id, types, updatedAttributes)
+            subscriptionService.getMatchingSubscriptions(id, updatedAttributes)
+                .filter {
+                    subscriptionService.isMatchingTypesQuery(
```
bobeal, line 39: no other solution than doing an extra query for types? :(

subscription-service/src/main/resources/db/migration/V0_19__expand_size_entity_info_type_solomn.sql
```diff
@@ -0,0 +1 @@
+alter table entity_info alter COLUMN "type" type varchar(1024);
```
bobeal, line 1: what is this doing here?

subscription-service/src/main/kotlin/com/egm/stellio/subscription/service/NotificationService.kt
gpoujol: otherwise I don't see how to retrieve the type of the entities part of a subscription to use the function `buildTypeQuery`

subscription-service/src/main/resources/db/migration/V0_19__expand_size_entity_info_type_solomn.sql
gpoujol: I update the size of the column because for some test the size is too small currently (128)

subscription-service/src/main/kotlin/com/egm/stellio/subscription/model/EntityTypeSelector.kt
gpoujol: I will change that

PR #884, opened 2023-03-19
refactor(subscription): improve and prepare geo-query handling for usage in other services

bobeal:


shared/src/main/kotlin/com/egm/stellio/shared/util/JsonLdUtils.kt
```diff
@@ -643,6 +643,9 @@ fun geoPropertyToWKT(jsonFragment: Map<String, Any>): Map<String, Any> {
             val geoAttribute = jsonFragment[geoProperty] as MutableMap<String, Any>
             val geoJsonAsString = geoAttribute[JSONLD_VALUE]
             val wktGeom = geoJsonToWkt(geoJsonAsString!! as Map<String, Any>)
+                .fold({
+                    throw BadRequestDataException(it.message)
```
gpoujol, line 647: it's not possible to do without the throw?

shared/src/main/kotlin/com/egm/stellio/shared/util/GeoQueryUtils.kt
```diff
@@ -1,61 +1,143 @@
...
-        georel = georel,
-        geometry = geometry,
-        coordinates = coordinates,
-        geoproperty = geoproperty
-    ).right()
+    // if at least one parameter is provided, the three must be provided for the geoquery to be valid
+    val notNullGeoParameters = listOfNotNull(georel, geometry, coordinates)
+    if (notNullGeoParameters.isEmpty())
+        null
+    else if (georel == null || geometry == null || coordinates == null)
```
gpoujol, line 68: This check is the same than above ?

shared/src/testFixtures/kotlin/com/egm/stellio/shared/util/NgsiLdTestUtils.kt
```diff
@@ -1,36 +1,32 @@
...
-import com.egm.stellio.shared.model.parseToNgsiLdAttributes
 
-fun parseGeoFragmentToPointGeoProperty(
+fun gimmeSimpleEntityWithGeoProperty(
     propertyKey: String,
     longitude: Double,
     latitude: Double
-): NgsiLdGeoProperty {
+): JsonLdEntity {
     val locationFragment =
```
gpoujol, line 11: it's not anymore a fragment of the location but rather an entity with just the location attribute now

shared/src/main/kotlin/com/egm/stellio/shared/util/JsonLdUtils.kt
bobeal: `JsonLdUtils` is unfortunately still mainly "exception-based". More specifically about this function, it is called from `doJsonLdExpansion` which already throws exception. Did not want to mix both concepts (nor to add more changes to this already quite big PR).

shared/src/main/kotlin/com/egm/stellio/shared/util/GeoQueryUtils.kt
bobeal: - 1st check: if empty, no geo parameter has been provided, geoquery is null
- 2nd check: not empty so at least one geo parameter has been provided. in this case, `georel`, `geometry` and `coordinates` must be provided

shared/src/testFixtures/kotlin/com/egm/stellio/shared/util/NgsiLdTestUtils.kt
bobeal: indeed

gpoujol (approved):

PR #885, opened 2023-03-20
feat: check Link header format

bobeal:
- also harmonize all web handlers wrt use of either effect


gpoujol:
I just have a question on the PR. Why now we can just write `either` whereas before we need to write `either<APIException, ResponseEntity<*>>` ?

PR #903, opened 2023-03-30
fix(temporal): inconsistent values for instanceId between specific column and instance payload

gpoujol:


search-service/src/main/resources/db/migration/V0_33__fix_instance_id_difference_in_its_column_and_payload.sql
```diff
@@ -0,0 +1,3 @@
+update attribute_instance
```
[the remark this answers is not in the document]
gpoujol: > that will fix newer attributes instances, but what about old ones?

The file `V0_27__fix_inconsistent_instance_id_values.sql` do it, no ?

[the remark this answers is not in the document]
bobeal: > > that will fix newer attributes instances, but what about old ones?
> 
> The file `V0_27__fix_inconsistent_instance_id_values.sql` do it, no ?

It is a migration file, it is run only once (fortunately!)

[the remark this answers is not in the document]
bobeal: > when I run this request in local (with dump of stellio-dev) I have `UPDATE 74494`

maybe but that was not my point. is it really fixing the problem?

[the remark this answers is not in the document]
gpoujol: the problem is not fixed here but when I add instance id in addition here 
```                
DO UPDATE SET value = :value, measured_value = :measured_value, payload = :payload,
                                    instance_id = :instance_id
```

bobeal:
one way to have less duplicate / boilerplate code would be to add a constructor in `AttributeInstance` that takes as input an `AttributeMetadata` (and other required parameters like the temporal entity attribute)?

gpoujol:
> one way to have less duplicate / boilerplate code would be to add a constructor in `AttributeInstance` that takes as input an `AttributeMetadata` (and other required parameters like the temporal entity attribute)?

yes I was thinking about doing something but I didn't know where to start I start from this point

search-service/src/main/resources/db/migration/V0_33__fix_instance_id_difference_in_its_column_and_payload.sql
```diff
@@ -0,0 +1,7 @@
...
+set instance_id = trim('"' FROM jsonb_path_query_first(payload, '$."https://uri.etsi.org/ngsi-ld/instanceId"[0]."@id"')::text)
+where instance_id != trim('"' FROM jsonb_path_query_first(payload, '$."https://uri.etsi.org/ngsi-ld/instanceId"[0]."@id"')::text);
+
+update attribute_instance
+set instance_id = trim('"' FROM jsonb_path_query_first(payload, '$."instanceId"')::text)
+where instance_id != trim('"' FROM jsonb_path_query_first(payload, '$."instanceId"')::text);
```
bobeal, line 7: - the `trim` could be avoided by using such expression: `jsonb_path_query_first(payload, '$."https://uri.etsi.org/ngsi-ld/instanceId"[0]')->>'@id'`
- this script is going to break a lot of the values in the `instance_id` column

search-service/src/main/resources/db/migration/V0_32__fix_instance_id_difference_in_its_column_and_payload.sql
```diff
@@ -0,0 +1,20 @@
+update attribute_instance
+set instance_id = jsonb_path_query_first(payload, '$."https://uri.etsi.org/ngsi-ld/instanceId"[0]')->>'@id'
+where instance_id != jsonb_path_query_first(payload, '$."https://uri.etsi.org/ngsi-ld/instanceId"[0]')->>'@id';
```
bobeal, line 3: as discussed, please add a comment about this (and the one below)

search-service/src/main/resources/db/migration/V0_32__fix_instance_id_difference_in_its_column_and_payload.sql
```diff
@@ -0,0 +1,20 @@
...
+
+update attribute_instance
+set instance_id = jsonb_path_query_first(payload, '$')->>'instanceId'
+where instance_id != jsonb_path_query_first(payload, '$')->>'instanceId';
+
+update attribute_instance_audit
```
bobeal, line 9: also explain why this

search-service/src/test/kotlin/com/egm/stellio/search/service/AttributeInstanceServiceTests.kt
```diff
@@ -588,15 +612,15 @@ class AttributeInstanceServiceTests : WithTimescaleContainer, WithKafkaContainer
...
+            attributeMetadata,
             attributeValues
         )
 
         verify {
             attributeInstanceService["create"](
                 match<AttributeInstance> {
                     it.time.toString() == "2015-10-18T11:20:30.000001Z" &&
-                        it.value == "false" &&
+                        it.value == false.toString() &&
```
bobeal, line 623: better keep the previous version (you want to check it has been transformed the way you expect it to)

search-service/src/main/kotlin/com/egm/stellio/search/model/AttributeInstance.kt
```diff
@@ -28,41 +29,69 @@ data class AttributeInstance private constructor(
...
-                parsedPayload.putIfAbsent(
-                    NGSILD_MODIFIED_AT_PROPERTY,
-                    buildNonReifiedDateTime(it)
-                )
-            }
+        ): AttributeInstance = AttributeInstance(
+            temporalEntityAttribute = temporalEntityAttribute,
+            instanceId = instanceId,
+            timeProperty = timeProperty,
+            time = (time ?: attributeMetadata.observedAt)!!,
```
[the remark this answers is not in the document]
gpoujol: what is better: return an error that no date has been entered or when no date has been entered used the date at the moment ?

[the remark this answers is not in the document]
bobeal: the best option is when the developer calling this constructor can't be wrong

search-service/src/main/resources/db/migration/V0_32__fix_instance_id_difference_in_its_column_and_payload.sql
```diff
@@ -1,11 +1,18 @@
+-- this sql query allows you to modify the instanceId of the column if it is different from the one contained in the payload
+-- this sql query works for payloads in expand format
+-- if the key of the json path does not exist it returns a null
+-- and the postgres operator returns false if one of the two parameters is null (https://www.postgresql.org/docs/current/functions-comparison.html)
```
bobeal, line 4: that is not true (fortunately for the script)

search-service/src/main/resources/db/migration/V0_32__fix_instance_id_difference_in_its_column_and_payload.sql
```diff
@@ -1,11 +1,18 @@
...
 update attribute_instance
 set instance_id = jsonb_path_query_first(payload, '$."https://uri.etsi.org/ngsi-ld/instanceId"[0]')->>'@id'
 where instance_id != jsonb_path_query_first(payload, '$."https://uri.etsi.org/ngsi-ld/instanceId"[0]')->>'@id';
 
+-- this sql query is the same as above but for non expandable formats (version before v2)
```
bobeal, line 9: they are not expanded, but they are expandable

search-service/src/main/kotlin/com/egm/stellio/search/service/AttributeInstanceService.kt
```diff
@@ -103,6 +103,7 @@ class AttributeInstanceService(
     ): Either<APIException, Unit> {
         val attributeInstance = AttributeInstance(
             temporalEntityAttribute = temporalEntityAttributeUuid,
+            time = attributeMetadata.observedAt!!,
```
bobeal, line 106: any hope to remove the last two `!!`?

gpoujol: the `observedAt` contained in `attributemetadata` is a `ZoneDateTime?`. I don't see how else to do it

search-service/src/main/resources/db/migration/V0_32__fix_instance_id_difference_in_its_column_and_payload.sql
```diff
@@ -0,0 +1,26 @@
+-- we modify the instancesId of the column if it is different from the one is contained in the payload
+-- this sql query works for expanded payload
+-- jsonb_path_query_first returns null if the key doesn't exist
+-- and the operator "!=" returns null (not true or false) when either input is null (https://www.postgresql.org/docs/current/functions-comparison.html)
```
bobeal, line 4:
```suggestion
-- modify the value of the instance_id column if different from the instance_id value contained in the payload
-- jsonb_path_query_first returns null if the key doesn't exist
-- and the operator "!=" returns null (not true or false) when either input is null (https://www.postgresql.org/docs/current/functions-comparison.html)
-- this query works for expanded payloads
```

search-service/src/main/resources/db/migration/V0_32__fix_instance_id_difference_in_its_column_and_payload.sql
```diff
@@ -0,0 +1,26 @@
...
+update attribute_instance
+set instance_id = jsonb_path_query_first(payload, '$."https://uri.etsi.org/ngsi-ld/instanceId"[0]')->>'@id'
+where instance_id != jsonb_path_query_first(payload, '$."https://uri.etsi.org/ngsi-ld/instanceId"[0]')->>'@id';
+
+-- this sql query is the same as above but for not expanded payload (version before v2)
```
bobeal, line 9:
```suggestion
-- same query as above but for compacted payloads (version before v2)
```

search-service/src/main/resources/db/migration/V0_32__fix_instance_id_difference_in_its_column_and_payload.sql
```diff
@@ -0,0 +1,26 @@
...
+update attribute_instance
+set instance_id = jsonb_path_query_first(payload, '$')->>'instanceId'
+where instance_id != jsonb_path_query_first(payload, '$')->>'instanceId';
+
+-- Currently the two attributes instances tables have the same instanceId. We fix this to only have different instanceId
```
bobeal, line 14:
```suggestion
-- the two attributes instances tables have the same instanceId for some updates of attributes
-- to prevent from any inconsistency, reset all the values of instanceId in audit table
```

bobeal (approved):

bobeal:
when merging, be sure to use the updated title for the PR

bobeal (approved):

PR #906, opened 2023-03-31
fix/expand payload for both attribute instance table

gpoujol:


search-service/src/main/kotlin/db/migration/V0_33__Expand_attribute_instance_payload.kt
```diff
@@ -0,0 +1,69 @@
...
+    private lateinit var jdbcTemplate: JdbcTemplate
+
+    override fun migrate(context: Context) {
+        jdbcTemplate = JdbcTemplate(SingleConnectionDataSource(context.connection, true))
+
+        queryAttributeUpdateByPage("attribute_instance", LIMIT, OFFSET)
+        queryAttributeUpdateByPage("attribute_instance_audit", LIMIT, OFFSET)
+    }
+
+    private fun queryAttributeUpdateByPage(tableName: String, limit: Int, offset: Int) {
```
bobeal, line 32: when does it stop?

gpoujol: if `resultSet` retunrs no data, nothing happens then so it stops at this moment. no?

bobeal: I just stopped the build in Jenkins that was running for 4 days so no

bobeal:
We can't run it in a migration script since it will take too much (unnecessary) time. So we have first to find a way to run it (once) in a non blocking way.

PR #914, opened 2023-04-15
refactor(common): NGSI-LD entity and attributes validation and building

bobeal:
- use extension functions to build NgsiLdXXX objects
- cleanup validation in NgsiLdXXX classes
- switch to Either-based programming model for NgsiLdXXX objects
- introduce more specialised expandAttribute(s) functions
- a lot of renaming to use expandXXX based names
- use as much as possible the ExpandedXXX type aliases


search-service/src/main/kotlin/com/egm/stellio/search/listener/ObservationEventListener.kt
```diff
@@ -99,18 +101,20 @@ class ObservationEventListener(
...
 
-        val ngsiLdAttributes = parseToNgsiLdAttributes(expandedPayload)
-        return entityPayloadService.appendAttributes(
+        val ngsiLdAttribute = expandedAttribute.toNgsiLdAttribute().bind()
+        entityPayloadService.appendAttributes(
             observationEvent.entityId,
-            ngsiLdAttributes,
-            expandedPayload,
+            listOf(ngsiLdAttribute),
+            expandedAttribute.toExpandedAttributes(),
```
gpoujol, line 117: Maybe use a variable to expandAttributes to avoid having to do expansion two times ?

PR #925, opened 2023-04-27
feat(devops): updated config for deployment into Kubernetes

agaldemas-eridanis:
## Proposed changes

new yaml files to deploy stellio under kubernetes,

## Types of changes

- Postgresql/Timescale is deployed as a StatefulSet, better than deployment for a database
- added a way to deploy a json-ld context server, with @context files you need, through ConfigMap

## Further comments

not tested on minikube, 
but on several different kubernetes versions (1.20 to 1.24), with Rancher RKE, RKE2, or OVH managed clusters, so it shall be OK with minikube too !


agaldemas-eridanis:
it's never too late...

bobeal (approved):

PR #937, opened 2023-05-26
refactor: remove event for subscription and notification

gpoujol:


search-service/src/main/resources/db/migration/V0_34__delete_all_subscription_and_notification_entity.sql
```diff
@@ -0,0 +1,3 @@
+delete from entity_payload
+where types && array['https://uri.etsi.org/ngsi-ld/Subscription']
+or types && array['https://uri.etsi.org/ngsi-ld/Notification'];
```
bobeal, line 3: also remove row in `temporal_entity_attribute` and `attribute_instance_xxx` tables (there should not be any rows in `entity_access_rights` but you can add a delete to be safe)

bobeal (approved):

PR #947, opened 2023-06-13
feat: implement merge entity endpoint - 6.5.3.4

gpoujol:


search-service/config/detekt/baseline.xml
```diff
@@ -30,5 +31,6 @@
     <ID>TooManyFunctions:TemporalEntityAttributeService.kt$TemporalEntityAttributeService</ID>
     <ID>UtilityClassWithPublicConstructor:ParameterizedTests.kt$ParameterizedTests</ID>
     <ID>UtilityClassWithPublicConstructor:QueryParameterizedTests.kt$QueryParameterizedTests</ID>
+    <ID>Wrapping:TemporalEntityAttributeService.kt$TemporalEntityAttributeService$(</ID>
```
bobeal, line 34: ?

search-service/src/main/kotlin/com/egm/stellio/search/web/EntityHandler.kt
```diff
@@ -82,6 +83,72 @@ class EntityHandler(
...
+        val sub = getSubFromSecurityContext()
+
+        entityPayloadService.checkEntityExistence(entityUri).bind()
+
+        val body = requestBody.awaitFirst().deserializeAsMap()
+            .checkNamesAreNgsiLdSupported().bind()
+            .checkContentIsNgsiLdSupported().bind()
+        val observedAt = options.getFirst(QUERY_PARAM_OPTIONS_OBSERVEDAT_VALUE)
+            ?.parseTimeParameter("'observedAt' parameter is not a valid date")
+            ?.getOrElse { return@either BadRequestDataException(it).left().bind<ResponseEntity<*>>() }
```
bobeal, line 106: the `observedAt` parameter is optional

search-service/src/main/kotlin/com/egm/stellio/search/web/EntityHandler.kt
```diff
@@ -82,6 +83,72 @@ class EntityHandler(
...
+
+        val contexts = checkAndGetContext(httpHeaders, body).bind()
+        val expandedAttributes = expandAttributes(body, contexts)
+        val (typeAttr, otherAttrs) = expandedAttributes.toList().partition { it.first == JsonLdUtils.JSONLD_TYPE }
+        val ngsiLdAttributes = otherAttrs.toMap().toNgsiLdAttributes().bind()
+
+        authorizationService.userCanUpdateEntity(entityUri, sub).bind()
```
bobeal, line 113: maybe it could be done sooner

search-service/src/main/kotlin/com/egm/stellio/search/service/EntityPayloadService.kt
```diff
@@ -103,6 +103,47 @@ class EntityPayloadService(
...
+        sub: Sub?
+    ): Either<APIException, UpdateResult> = either {
+        val createdAt = ZonedDateTime.now(ZoneOffset.UTC)
+        logger.debug("Merging entity {}", entityId)
+
+        val temporalEntityAttributes = temporalEntityAttributeService.getForEntity(entityId, emptySet())
+        temporalEntityAttributes.filter {
+            !ngsiLdAttributes.map { it.name }.contains(it.attributeName)
+        }.forEach {
+            temporalEntityAttributeService.deleteTemporalAttribute(
```
bobeal, line 121: I don't get why you are deleting attributes

search-service/src/main/kotlin/com/egm/stellio/search/service/TemporalEntityAttributeService.kt
```diff
@@ -714,6 +715,75 @@ class TemporalEntityAttributeService(
...
+        expandedAttributes: ExpandedAttributes,
+        teas: List<TemporalEntityAttribute>,
+        createdAt: ZonedDateTime,
+        observedAt: ZonedDateTime?,
+        sub: Sub?
+    ): Either<APIException, UpdateResult> =
+        either {
+            val attributeInstances = ngsiLdAttributes.flatOnInstances()
+            attributeInstances.parTraverseEither { (ngsiLdAttribute, ngsiLdAttributeInstance) ->
+                logger.debug("Appending attribute {} in entity {}", ngsiLdAttribute.name, entityUri)
```
bobeal, line 731: you are not appending (at least not yet)

search-service/src/main/kotlin/com/egm/stellio/search/service/EntityPayloadService.kt
gpoujol: the description in the spec says that `This operation allows modification of an existing NGSI-LD Entity aligning to the JSON Merge Patch processing rules defined in IETF RFC 7396 [16] by adding new Attributes (Properties or Relationships) or modifying or deleting existing Attributes associated with an existing Entity.`. So for me, I understood that it is necessary to add attributes when they do not exist, update them when they exist and delete them if they are no longer present

bobeal: the removal is done via the use of NGSI-LD Null (see 5.6.17 - Merge Entity), that we'll implement later

search-service/src/test/kotlin/com/egm/stellio/search/web/EntityHandlerTests.kt
```diff
@@ -1745,6 +1745,135 @@ class EntityHandlerTests {
...
+
+        webClient.patch()
+            .uri("/ngsi-ld/v1/entities/$entityId")
+            .bodyValue(jsonLdFile)
+            .exchange()
+            .expectStatus().isNotFound
+            .expectBody().json(
+                "{\"type\":\"https://uri.etsi.org/ngsi-ld/errors/ResourceNotFound\"," +
+                    "\"title\":\"The referred resource has not been found\"," +
+                    "\"detail\":\"Entity $entityId was not found\"}"
```
bobeal, line 1823: use the multi-line notation for readability

search-service/src/test/kotlin/com/egm/stellio/search/web/EntityHandlerTests.kt
```diff
@@ -1745,6 +1745,135 @@ class EntityHandlerTests {
...
+                any(),
+                eq(updateResult),
+                true,
+                eq(listOf(AQUAC_COMPOUND_CONTEXT))
+            )
+        }
+    }
+
+    @Test
+    fun `merge entity should return a 404 if entityId if entity does not exist`() {
```
bobeal, line 1807: "if entityId if entity..."

search-service/src/main/kotlin/com/egm/stellio/search/web/EntityHandler.kt
bobeal: the observedAt info is conveyed in a specific `observedAt` request parameter (see Table 6.5.3.4-1). a good opportunity to add a test.

search-service/src/main/kotlin/com/egm/stellio/search/service/EntityPayloadService.kt
```diff
@@ -103,6 +103,36 @@ class EntityPayloadService(
...
 
+    @Transactional
+    suspend fun mergeEntity(
+        entityId: URI,
+        ngsiLdAttributes: List<NgsiLdAttribute>,
+        expandedAttributes: ExpandedAttributes,
+        observedAt: ZonedDateTime?,
+        sub: Sub?
+    ): Either<APIException, UpdateResult> = either {
+        val createdAt = ZonedDateTime.now(ZoneOffset.UTC)
```
bobeal, line 114: look more like a `modifiedAt` than a `createdAt`

bobeal: also you could use the `ngsildDateTime()` function?

search-service/src/main/kotlin/com/egm/stellio/search/service/EntityPayloadService.kt
```diff
@@ -103,6 +103,36 @@ class EntityPayloadService(
...
+        logger.debug("Merging entity {}", entityId)
+
+        val temporalEntityAttributes = temporalEntityAttributeService.getForEntity(entityId, emptySet())
```
bobeal, line 117: seems like an expensive operation to get data you'll often don't need. and in any case, it is better to do it in TemporalEntityAttributeService since it is part of the logic of merging entity attributes (EntityPayloadService does not care about this)

search-service/src/main/kotlin/com/egm/stellio/search/service/TemporalEntityAttributeService.kt
```diff
@@ -714,6 +768,78 @@ class TemporalEntityAttributeService(
...
+                logger.debug("Merging attribute {} in entity {}", ngsiLdAttribute.name, entityUri)
+                val currentTea = teas.firstOrNull {
+                    it.attributeName == ngsiLdAttribute.name && it.datasetId == ngsiLdAttributeInstance.datasetId
+                }
+                var attributeMetadata = ngsiLdAttributeInstance.toTemporalAttributeMetadata().bind()
+                var attributePayload = getAttributeFromExpandedAttributes(
+                    expandedAttributes,
+                    ngsiLdAttribute.name,
+                    ngsiLdAttributeInstance.datasetId
+                )!!
```
bobeal, line 793: `var` 😢

search-service/src/main/kotlin/com/egm/stellio/search/service/TemporalEntityAttributeService.kt
```diff
@@ -714,6 +768,78 @@ class TemporalEntityAttributeService(
...
+
+                if (
+                    observedAt != null &&
+                    currentTea?.payload.toString().contains(NGSILD_OBSERVED_AT_PROPERTY) &&
+                    !attributePayload.containsKey(NGSILD_OBSERVED_AT_PROPERTY)
+                ) {
+                    attributePayload =
+                        attributePayload.plus(NGSILD_OBSERVED_AT_PROPERTY to buildNonReifiedDateTime(observedAt))
+                    attributeMetadata = attributeMetadata.copy(observedAt = observedAt)
+                }
```
bobeal, line 803: since it is specific to the merge of an existing attribute, maybe it could be moved there?

search-service/src/main/kotlin/com/egm/stellio/search/service/TemporalEntityAttributeService.kt
```diff
@@ -248,6 +249,59 @@ class TemporalEntityAttributeService(
             }
         }
 
+    @Transactional
+    suspend fun mergeAttribute(
```
bobeal, line 253: some of the logic here could be extracted and mutualized with the partial attribute operation (which is very similar... even identical with the exception of the default `observedAt`?)

search-service/src/test/kotlin/com/egm/stellio/search/service/TemporalEntityAttributeServiceTests.kt
```diff
@@ -248,7 +249,7 @@ class TemporalEntityAttributeServiceTests : WithTimescaleContainer, WithKafkaCon
         temporalEntityAttributeService.createEntityTemporalReferences(rawEntity, listOf(APIC_COMPOUND_CONTEXT))
             .shouldSucceed()
 
-        val createdAt = ZonedDateTime.now()
+        val createdAt = ngsiLdDateTime()
```
bobeal, line 252: you are modifying it here... but not in the tests you are adding (see for instance line 300)...

search-service/src/test/resources/ngsild/fragments/beehive_incoming_property.json
```diff
@@ -0,0 +1,11 @@
+{
```
bobeal, line 1: does not seem to be used?

bobeal:
First feedback (I've still to look at the tests)

search-service/src/main/kotlin/com/egm/stellio/search/service/EntityPayloadService.kt
gpoujol: It's a modifiedAt when we modify attributes which exist but when we create new attributes it's createdAt

search-service/src/test/resources/ngsild/fragments/beehive_incoming_property.json
gpoujol: yes indeed

search-service/src/main/kotlin/com/egm/stellio/search/service/EntityPayloadService.kt
bobeal: for the replace entity endpoint you implemented just before, it is called `replacedAt`. so I would suggest to call it `mergedAt` here (it avoids confusions)

gpoujol: ok i will change that

search-service/src/test/kotlin/com/egm/stellio/search/service/TemporalEntityAttributeServiceTests.kt
```diff
@@ -282,6 +283,164 @@ class TemporalEntityAttributeServiceTests : WithTimescaleContainer, WithKafkaCon
...
+            .shouldSucceed()
+
+        val temporalEntityAttribute = temporalEntityAttributeService.getForEntityAndAttribute(
+            beehiveTestCId,
+            INCOMING_PROPERTY
+        ).shouldSucceedAndResult()
+
+        val mergedAt = ngsiLdDateTime()
+        val observedAt = ngsiLdDateTime()
+        val newProperty = loadSampleData("fragments/beehive_mergeAttribute.json")
```
bobeal, line 302: If I got it correctly, it is not a new property but one to be merged

search-service/src/test/kotlin/com/egm/stellio/search/service/TemporalEntityAttributeServiceTests.kt
```diff
@@ -282,6 +283,164 @@ class TemporalEntityAttributeServiceTests : WithTimescaleContainer, WithKafkaCon
...
+        )
+
+        temporalEntityAttributeService.getForEntityAndAttribute(
+            beehiveTestCId,
+            INCOMING_PROPERTY
+        ).shouldSucceedWith {
+            it.attributeType == TemporalEntityAttribute.AttributeType.Property &&
+                it.attributeValueType == TemporalEntityAttribute.AttributeValueType.STRING &&
+                it.entityId == beehiveTestCId &&
+                it.payload.asString() == serializeObject(expandedAttribute) &&
```
bobeal, line 329: I'm a bit surprised with this check. Since the original entity has a `incoming` property, the resulting payload should be the merge of the original and the one you are using in the merge operation (and not the `expandedAttribute`)

search-service/src/test/kotlin/com/egm/stellio/search/service/TemporalEntityAttributeServiceTests.kt
```diff
@@ -282,6 +283,164 @@ class TemporalEntityAttributeServiceTests : WithTimescaleContainer, WithKafkaCon
...
+        // 4 when we create entity attributes and 2 when we merge entity attributes
+        coVerify(exactly = 6) {
+            attributeInstanceService.create(
+                match { it.timeProperty == AttributeInstance.TemporalProperty.CREATED_AT }
+            )
+        }
+    }
+
+    @Test
+    fun `it should merge an entity attributes with observedAt parameter`() = runTest {
```
bobeal, line 399: if you just want to test the behavior when an `observedAt` parameter is provided, better to use more focused test fixture. there are a lot of attributes in `beehive_mergeAttributes_without_observedAt.json` but it seems only the `incoming` property is tested, which makes the whole thing hard to understand.

maybe use fixture data with one property with an `observedAt` and another one without and test the two resulting TEAs?

search-service/src/main/kotlin/com/egm/stellio/search/service/TemporalEntityAttributeService.kt
bobeal: just gone through the spec and expected behaviors are actually not the same for partial attribute update (5.5.8) and merge entity with an existing attribute (5.5.12)

search-service/src/test/kotlin/com/egm/stellio/search/service/TemporalEntityAttributeServiceTests.kt
gpoujol: this is because the payload that we retrieve is that of the `incoming` attribute

bobeal: yes... but I don't see how it answers my point...

PR #977, opened 2023-08-07
feat(authz): get a list of all users (reserved to stellio admins)

bobeal:


junjou (approved):
I can see all important unit tests are implemented so... LGTM !

PR #1098, opened 2024-02-14
feat(subscription): add support for sysAttrs to notification params

ranim-n:


subscription-service/src/test/kotlin/com/egm/stellio/subscription/service/NotificationServiceTests.kt
```diff
@@ -404,4 +422,120 @@ class NotificationServiceTests {
...
+                    "type" to "Apiary",
+                    "createdAt" to  "2024-02-13T18:15:00Z",
+                    "modifiedAt" to "2024-02-13T18:16:00Z",
+                    "name" to mapOf(
+                        "type" to "Property",
+                        "value" to "ApiarySophia",
+                        "createdAt" to "2024-02-13T18:15:00Z",
+                        "modifiedAt" to "2024-02-13T18:16:00Z"
+                    ),
+                    "@context" to APIC_COMPOUND_CONTEXT
```
bobeal, line 499: you should not have all these things in a subscription. in the subscription, you only say what you want to be notified about, you don't put an entity in the subscription (see 5.2.12).

subscription-service/src/test/kotlin/com/egm/stellio/subscription/service/NotificationServiceTests.kt
```diff
@@ -404,4 +422,120 @@ class NotificationServiceTests {
 
         verify(1, postRequestedFor(urlPathEqualTo("/notification")))
     }
+
+    @Test
+    fun `it should notify the subscriber and return entities without system attributes if sysAttrs is false`() = runTest {
+
+        val payload = mapOf(
```
bobeal, line 429: btw, when don't you use the `gimmeRawSubscription` helper function as it is done in other tests?

ranim-n: I thought a subscription must be created with an entity in this case and that method creates ones without entities, that's why I didn't use it

ranim-n: in this case it can be used , since sysAttrs is by default false , but when it needs to be set to true I can't use it.

bobeal: you can use it and then create a copy a subscription, overriding what you need to overwrite (it is already done in many other tests in the class)

PR #1102, opened 2024-02-21
fix(subscription): notification when last notification is null and throttling not null

ranim-n:


subscription-service/src/test/kotlin/com/egm/stellio/subscription/service/SubscriptionServiceTests.kt
```diff
@@ -1171,6 +1171,32 @@ class SubscriptionServiceTests : WithTimescaleContainer, WithKafkaContainer {
         }
     }
 
+    @Test
+    fun `it should return a subscription if last_notification is null`() = runTest {
```
bobeal, line 1175:
```suggestion
    fun `it should return a subscription if throttling is not null and last_notification is null`() = runTest {
```

bobeal (approved):

PR #1115, opened 2024-03-20
fix(core): do not force existing types to be present in payload when updating or appending types

ranim-n:


search-service/src/test/kotlin/com/egm/stellio/search/service/EntityPayloadServiceTests.kt
```diff
@@ -586,16 +586,14 @@ class EntityPayloadServiceTests : WithTimescaleContainer, WithKafkaContainer {
...
+        entityPayloadService.retrieve(entity01Uri)
             .shouldSucceedWith {
-                assertFalse(it.isSuccessful())
-                assertEquals(1, it.notUpdated.size)
-                val notUpdatedDetails = it.notUpdated[0]
-                assertEquals(JSONLD_TYPE, notUpdatedDetails.attributeName)
                 assertEquals(
-                    "A type cannot be removed from an entity: [$BEEHIVE_TYPE] have been removed",
-                    notUpdatedDetails.reason
+                    listOf(BEEHIVE_TYPE, APIARY_TYPE),
```
bobeal, line 595: the SQL request is actually working:

image

the `types` column is correctly updated, it is the `type` in the payload that is not updated (see lines 427-431 in `EntityPayloadService.kt`)

search-service/src/main/kotlin/com/egm/stellio/search/service/EntityPayloadService.kt
```diff
@@ -424,18 +424,6 @@ class EntityPayloadService(
...
-                        UpdateAttributeResult(
-                            attributeName = JSONLD_TYPE,
-                            updateOperationResult = UpdateOperationResult.FAILED,
-                            errorMessage = "A type cannot be removed from an entity: $removedTypes have been removed"
-                        )
-                    )
-                )
-            }
 
             val updatedPayload = entityPayload.payload.deserializeExpandedPayload()
```
bobeal, line 428: since `payload` has also to be updated, it seems there is no choice but to retrieve the entity from the DB

bobeal:
ready for review? (if so, change the status of the PR)

bobeal:
Set a more meaningful name to the PR (and you can merge after)

PR #1142, opened 2024-04-24
fix: error code when requesting a pagination limit that surpasses the configured value 

ranim-n:


.env
```diff
@@ -22,3 +22,7 @@ SUBSCRIPTION_STELLIO_URL=http://localhost:8080
 APPLICATION_TENANTS_0_ISSUER=https://sso.eglobalmark.com/auth/realms/stellio
 APPLICATION_TENANTS_0_NAME=urn:ngsi-ld:tenant:default
 APPLICATION_TENANTS_0_DBSCHEMA=public
+
+# Pagination config for query resources endpoints
+LIMIT_DEFAULT=30
+LIMIT_MAX=100
```
bobeal, line 28: give them the same name as in docker-compose (and in the properties file in the code). no need to add some complexity learning a new name for something that already exists.

docker-compose.yml
```diff
@@ -60,6 +60,8 @@ services:
       - APPLICATION_TENANTS_0_ISSUER=${APPLICATION_TENANTS_0_ISSUER}
       - APPLICATION_TENANTS_0_NAME=${APPLICATION_TENANTS_0_NAME}
       - APPLICATION_TENANTS_0_DBSCHEMA=${APPLICATION_TENANTS_0_DBSCHEMA}
+      - APPLICATION_PAGINATION_LIMITDEFAULT=${LIMIT_DEFAULT}
```
bobeal, line 63: I'm not sure this will work without the `-` between limit and default

shared/src/main/kotlin/com/egm/stellio/shared/model/ErrorResponse.kt
```diff
@@ -40,10 +40,9 @@ data class TooComplexQueryResponse(override val detail: String) : ErrorResponse(
 
 data class TooManyResultsResponse(override val detail: String) : ErrorResponse(
     ErrorType.TOO_MANY_RESULTS.type,
-    """
-    The query associated to the operation is producing so many results that can exhaust client or server resources. 
-    It should be made more restrictive
-    """,
+    "The query associated to the operation is producing so many results " +
+        "that can exhaust client or server resources. " +
+        "It should be made more restrictive",
```
bobeal, line 45: use multi-line string, as it was before. easier to read, no?

docker-compose.yml
ranim-n: in the documentation it is said that `-` must be removed and replaced like that

shared/src/main/kotlin/com/egm/stellio/shared/model/ErrorResponse.kt
ranim-n: with the multi-line string the tests fail because the json string can't match. I tried \n but it still failed

bobeal: indeed, you would have to regex replace the `\n` in the value for the final result to be a single-line string (which is what is expected in the API response). not a problem but it adds not necessary overhead, keep it as it is.

docker-compose.yml
bobeal: indeed, it will also work (not a "must be removed", look at other definition in the same docker compose file).

ranim-n: in the docker compose file there are no other examples of properties with `-` to bind with. Should I add the` _ ` then ?

bobeal: `SUBSCRIPTION_ENTITY-SERVICE-URL` maybe ?

bobeal (approved):

PR #1146, opened 2024-04-27
feat(core): align behavior for update entity attributes (1.6+ compliance)

bobeal:


search-service/src/main/kotlin/com/egm/stellio/search/service/TemporalEntityAttributeService.kt
```diff
@@ -593,12 +593,12 @@ class TemporalEntityAttributeService(
             val currentTea =
                 getForEntityAndAttribute(entityUri, ngsiLdAttribute.name, ngsiLdAttributeInstance.datasetId)
                     .fold({ null }, { it })
+            val attributeMetadata = ngsiLdAttributeInstance.toTemporalAttributeMetadata().bind()
+            val attributePayload = expandedAttributes.getAttributeFromExpandedAttributes(
+                ngsiLdAttribute.name,
+                ngsiLdAttributeInstance.datasetId
+            )!!
             if (currentTea != null) {
```
ranim-n, line 601: why did you place them outside the if clause ?

search-service/src/main/kotlin/com/egm/stellio/search/service/TemporalEntityAttributeService.kt
```diff
@@ -615,17 +615,21 @@ class TemporalEntityAttributeService(
...
             } else {
-                val message = if (ngsiLdAttributeInstance.datasetId != null)
-                    "Attribute (datasetId: ${ngsiLdAttributeInstance.datasetId}) does not exist"
-                else
-                    "Attribute (default instance) does not exist"
-                logger.info(message)
-                UpdateAttributeResult(
+                addAttribute(
+                    entityUri,
                     ngsiLdAttribute.name,
```
ranim-n, line 620: is this to treate the case where a temporal entity doesn't have any instances of this attribute yet ?

search-service/src/main/kotlin/com/egm/stellio/search/service/TemporalEntityAttributeService.kt
bobeal: because I need them in both clauses

search-service/src/main/kotlin/com/egm/stellio/search/service/TemporalEntityAttributeService.kt
bobeal: in 1.6, new behavior states that if the attribute does not exist, it has to be added (before it had to be ignored)

ranim-n (approved):

PR #1149, opened 2024-04-28
feat: merge service account id to subject id

bobeal:
- if a client has service account enabled, the sub presented in the JWT is the service account id
- the client id is never used for authentication purposes and can then be dropped (a client cannot authenticate by itself if service account is not enabled)


ranim-n (commented):
what is exactly a client's service account ?

bobeal:
> what is exactly a client's service account ?

`A service account is a type of client that is able to obtain tokens on its own behalf.` (https://www.keycloak.org/docs/latest/securing_apps/#terminology)

ranim-n (approved):

PR #1160, opened 2024-05-15
fix: temporal entity queries with lastN parameter and aggregation period duration 

ranim-n:


bobeal:
@ranim-n I did some modifications, seems good now.

search-service/src/test/kotlin/com/egm/stellio/search/service/AttributeInstanceServiceTests.kt
```diff
@@ -366,21 +366,27 @@ class AttributeInstanceServiceTests : WithTimescaleContainer, WithKafkaContainer
...
                 timerel = TemporalQuery.Timerel.BEFORE,
-                timeAt = now.plusHours(1),
+                timeAt = now,
                 aggrPeriodDuration = "PT1S",
                 aggrMethods = listOf(TemporalQuery.Aggregate.SUM),
                 lastN = 5
-            )
+            ),
+            withAggregatedValues = true
         )
```
ranim-n, line 390: should we also test the case where `lastN` is added but `withAggregatedValues = false` (basically we specify an aggregation method but we don't sepcify whether we want an aggregated or temporal values)

bobeal: if there is not one yet, yes.

ranim-n: that case is what causes an issue, the test fails if `withAggregatedValues = true` is removed

search-service/src/test/kotlin/com/egm/stellio/search/service/AttributeInstanceServiceTests.kt
```diff
@@ -365,6 +365,64 @@ class AttributeInstanceServiceTests : WithTimescaleContainer, WithKafkaContainer
...
+                        time = now.minusSeconds(index.toLong())
+                    )
+            attributeInstanceService.create(attributeInstance)
+        }
+
+        val temporalEntitiesQuery = gimmeTemporalEntitiesQuery(
+            TemporalQuery(
+                timerel = TemporalQuery.Timerel.BEFORE,
+                timeAt = now,
+                aggrMethods = listOf(TemporalQuery.Aggregate.SUM),
```
bobeal, line 385: don't set `aggrMethods` if you are not doing an aggregation

search-service/src/test/kotlin/com/egm/stellio/search/service/AttributeInstanceServiceTests.kt
```diff
@@ -365,6 +365,64 @@ class AttributeInstanceServiceTests : WithTimescaleContainer, WithKafkaContainer
             }
     }
 
+    @Test
+    fun `it should only return the last n instances without aggregated or temporal values specified`() = runTest {
```
bobeal, line 369: I don't really see the difference between this test and the one just above (`it should only return the last n instances asked in the temporal query`)

ranim-n: here I removed` withAggregatedValues` and `aggrPeriodDuration`

bobeal: in the test above, `withAggregatedValues` is also not provided (and so is false), nor `aggrPeriodDuration`

ranim-n: an `aggrMethods` was added. But since you said it should also be removed then this test is unecessary

bobeal: if `withAggregatedValues` is false, `aggrMethods` does not make sense.

PR #1166, opened 2024-05-21
Batch Merge

thomasBousselin:
add batch merge


thomasBousselin:
@ranim-n

search-service/src/main/kotlin/com/egm/stellio/search/service/EntityOperationService.kt
```diff
@@ -234,4 +246,19 @@ class EntityOperationService(
             sub
         ).bind()
     }
+
+    @SuppressWarnings("UnusedParameter")
+    suspend fun mergeEntity(
```
bobeal, line 251: others (`replaceEntity`, ...) are wrapped into an `either` and do a `bind()`. but not sure it is actually needed in this case.

search-service/src/test/kotlin/com/egm/stellio/search/service/EntityOperationServiceTests.kt
```diff
@@ -382,4 +382,31 @@ class EntityOperationServiceTests {
         coVerify { entityPayloadService.deleteEntity(firstEntityURI) }
         coVerify { entityPayloadService.deleteEntity(secondEntityURI) }
     }
+
+    @Test
+    fun `batch merge should ask to merge attributes of entities`() = runTest {
+        coEvery {
```
ranim-n, line 388: why is the test called `...should ask to merge...` ?

thomasBousselin: Because it should call the merge function from entityPayloadHandler.
I can rename it "batch merge should call mergeEntity" to be more accurate.

PR #1199, opened 2024-07-23
feature: delete entities owned by a deleted user

ranim-n:


search-service/src/main/kotlin/com/egm/stellio/search/config/SearchProperties.kt
```diff
@@ -4,5 +4,6 @@ import org.springframework.boot.context.properties.ConfigurationProperties
 
 @ConfigurationProperties("search")
 data class SearchProperties(
-    val payloadMaxBodySize: Int
+    val payloadMaxBodySize: Int,
+    var deleteEntityByCreator: Boolean
```
thomasBousselin, line 8: Could be false by default. So you add the property is used only when you need the funcionnality

search-service/src/main/kotlin/com/egm/stellio/search/listener/IAMListener.kt
```diff
@@ -99,9 +105,19 @@ class IAMListener(
         tenantName: String,
         entityDeleteEvent: EntityDeleteEvent
     ): Either<APIException, Unit> = either {
+        val subjectType = SubjectType.valueOf(entityDeleteEvent.entityTypes.first().uppercase())
         mono {
             subjectReferentialService.delete(entityDeleteEvent.entityId.extractSub())
         }.writeContextAndSubscribe(tenantName, entityDeleteEvent)
+        if (searchProperties.deleteEntityByCreator && subjectType == SubjectType.USER) {
+            val entitiesIds = entityAccessRightsService.getEntitiesIdsOwnedBySubject(
+                entityDeleteEvent.entityId.extractSub()
```
thomasBousselin, line 114: I had difficutlies understanding this part maybe put```entityDeleteEvent.entityId.extractSub()``` in a deletedUserId variable, to clarify what it is.

search-service/src/main/kotlin/com/egm/stellio/search/authorization/EntityAccessRightsService.kt
```diff
@@ -321,6 +321,22 @@ class EntityAccessRightsService(
             }
     }
 
+    suspend fun getEntitiesIdsOwnedBySubject(
```
thomasBousselin, line 324: Why not directly do a deleteEntitiesOwnedBySubject?

search-service/src/main/resources/application.properties
```diff
@@ -13,3 +13,6 @@ server.port = 8083
 
 # Allow body size in payloads up to 2Mb
 search.payload-max-body-size = 2048000
+
+# Allow creator of an entity to delete it
+search.delete-entity-by-creator = false
```
thomasBousselin, line 18: Is there a name that explain what the variable do?
for example : 
- delete-entity-with-their-owner
- owner-cascade-delete
- other?

That way you can get rid off the comment to.

bobeal: - `owner-delete-cascade-entities`
- `on-owner-delete-cascade-entities`

search-service/src/main/kotlin/com/egm/stellio/search/authorization/EntityAccessRightsService.kt
ranim-n: the deletion doesn't happen unless you recieve an `EntityDeleteEvent` in `IAMListener`. in `IAMListener`  the user (`SubjectReferential` for better terms) is deleted then the entities created by that user are deleted.
We need to first get those entities using the user's access rights. 
This way the code is better organized with every service doing his part of the whole process and _getEntitiesIdsOwnedBySubject()_ can be used in other places.

search-service/src/main/kotlin/com/egm/stellio/search/config/SearchProperties.kt
ranim-n: in `search-service/src/main/resources/application.properties` it is `false` by default

search-service/src/main/kotlin/com/egm/stellio/search/authorization/EntityAccessRightsService.kt
bobeal: Agree

search-service/src/main/kotlin/com/egm/stellio/search/config/SearchProperties.kt
```diff
@@ -4,5 +4,6 @@ import org.springframework.boot.context.properties.ConfigurationProperties
 
 @ConfigurationProperties("search")
 data class SearchProperties(
-    val payloadMaxBodySize: Int
+    val payloadMaxBodySize: Int,
+    var ownerDeleteCascadeEntities: Boolean
```
bobeal, line 8: missing a prefix `on`? 😢

search-service/src/main/kotlin/com/egm/stellio/search/listener/IAMListener.kt
```diff
@@ -99,9 +105,20 @@ class IAMListener(
...
             subjectReferentialService.delete(entityDeleteEvent.entityId.extractSub())
         }.writeContextAndSubscribe(tenantName, entityDeleteEvent)
+        if (searchProperties.ownerDeleteCascadeEntities && subjectType == SubjectType.USER) {
+            val entitiesIds = entityAccessRightsService.getEntitiesIdsOwnedBySubject(
+                sub
+            )
+                .getOrNull()
+            entitiesIds?.forEach { entityId ->
+                entityPayloadService.deleteEntity(entityId)
+            }
```
bobeal, line 120: you could add a new `entityPayloadService.deleteEntities(...)` function

ranim-n: Batch Entity Delete is a seperate feature (see `5.6.10` ), it would be better to implement it in a seperate PR.

bobeal: this was not my point (and btw, batch entity delete is already implemented in Stellio).

my point was that, instead of calling `entityPayloadService.deleteEntity(entityId)` for each entity, add a new function `entityPayloadService.deleteEntities(...)` that accepts a list of entities

bobeal: btw, `entityOperationService.delete()` is already doing the job

bobeal (approved):

PR #1211, opened 2024-08-03
fix(core): error sending Kafka events for batch operations when many times the same entity id

bobeal:


search-service/src/test/kotlin/com/egm/stellio/search/service/EntityOperationServiceTests.kt
```diff
@@ -150,53 +153,72 @@ class EntityOperationServiceTests {
...
     @Test
-    fun `processEntities should count as error not processed attributes in entities`() =
-        runTest {
-            val updateResult = UpdateResult(
-                emptyList(),
-                listOf(
-                    NotUpdatedDetails("attribute#1", "reason"),
-                    NotUpdatedDetails("attribute#2", "reason")
-                )
+    fun `processEntities should count as error not processed attributes in entities`() = runTest {
```
ranim-n, line 177: the name of the test is a bit confusing. I'm not sure I understand what it should do

search-service/src/test/kotlin/com/egm/stellio/search/service/EntityOperationServiceTests.kt
```diff
@@ -297,21 +331,30 @@ class EntityOperationServiceTests {
...
         coEvery { entityPayloadService.deleteEntity(any()) } returns Unit.right()
         coEvery { authorizationService.removeRightsOnEntity(any()) } returns Unit.right()
+        coEvery { entityEventService.publishEntityDeleteEvent(any(), any()) } returns Job()
 
-        val batchOperationResult = entityOperationService.delete(setOf(firstEntityURI, secondEntityURI))
+        val batchOperationResult = entityOperationService.delete(
+            setOf(
+                mockkClass(EntityPayload::class) { every { entityId } returns firstEntityURI },
+                mockkClass(EntityPayload::class) { every { entityId } returns secondEntityURI },
+            ),
```
ranim-n, line 355: here : `every { entityId } returns firstEntityURI` and `every { entityId } returns secondEntityURI` is because in `entityOperationService.delete()` entities Ids are extracted to delete each entity ?

search-service/src/test/kotlin/com/egm/stellio/search/service/EntityOperationServiceTests.kt
bobeal: it was alredy there before, not adding unnecessary modifications

search-service/src/test/kotlin/com/egm/stellio/search/service/EntityOperationServiceTests.kt
bobeal: `entityOperationService.delete()` now receives a set of `EntityPayload` objects instead of entities ids and it reads the id of each entity so it needs to have some.

ranim-n (approved):

thomasBousselin:
Now the kafka event is triggered in the service. What else have you changed to to fix the issue?

PR #1219, opened 2024-08-14
feat: refactor entity and temporal handlers

bobeal:


search-service/src/test/kotlin/com/egm/stellio/search/entity/web/EntityOperationHandlerTests.kt
```diff
@@ -269,57 +225,14 @@ class EntityOperationHandlerTests {
         }
     }
 
-    @Test
```
bobeal: @thomasBousselin you removed some tests but they could / should still be tested in the handler (not enough rights, non existing entity in the batch)

thomasBousselin: We should however check those point in the entityServiceTest; If they do not exist yet.

bobeal: yes, but in the handler, it is important to check that it correctly accumulates the errors and successes for each entity that is in the batch.

thomasBousselin: After double checking, i think (line 137) the test "update batch entity should return a 207 if JSON-LD payload contains update errors" already check this case.

thomasBousselin:
Les deux test suites sont bien passé sans erreurs

search-service/src/main/kotlin/com/egm/stellio/search/authorization/listener/IAMListener.kt
```diff
@@ -114,19 +112,15 @@ class IAMListener(
         val subjectType = SubjectType.valueOf(entityDeleteEvent.entityTypes.first().uppercase())
         val sub = entityDeleteEvent.entityId.extractSub()
         mono {
+            // delete the entities owned by the user while the user still exists
+            // (if it no longer exists, it fails because of access rights checks)
+            if (searchProperties.onOwnerDeleteCascadeEntities && subjectType == SubjectType.USER) {
+                entityAccessRightsService.getEntitiesIdsOwnedBySubject(sub).getOrNull()?.forEach { entityId ->
+                    entityService.deleteEntity(entityId, sub)
+                }
```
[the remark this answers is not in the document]
bobeal: It is done in the deleteEntity function

ranim-n (approved):

PR #1245, opened 2024-09-28
feat: allow setting a specific timezone for aggregates

bobeal:
By default, bucket start times are aligned at 00:00:00UTC (see https://docs.timescale.com/use-timescale/latest/time-buckets/about-time-buckets/#timezones). This property allows to align them different timezones which is useful for weekly or monthly aggregates for instance.

### DIFF
--- a/search-service/src/test/kotlin/com/egm/stellio/search/temporal/service/AggregatedTemporalQueryServiceTests.kt
+++ b/search-service/src/test/kotlin/com/egm/stellio/search/temporal/service/AggregatedTemporalQueryServiceTests.kt
@@ -1,11 +1,14 @@
 package com.egm.stellio.search.temporal.service
 
+import com.egm.stellio.search.common.config.SearchProperties
 import com.egm.stellio.search.entity.model.Attribute
 import com.egm.stellio.search.entity.service.EntityAttributeService
 import com.egm.stellio.search.support.*
 import com.egm.stellio.search.temporal.model.*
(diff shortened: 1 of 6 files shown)


ranim-n (approved):

search-service/src/test/kotlin/com/egm/stellio/search/temporal/service/AggregatedTemporalQueryServiceTests.kt
```diff
@@ -39,10 +43,18 @@ class AggregatedTemporalQueryServiceTests : WithTimescaleContainer, WithKafkaCon
...
     private lateinit var r2dbcEntityTemplate: R2dbcEntityTemplate
 
+    @MockkBean(relaxed = true)
+    private lateinit var searchProperties: SearchProperties
+
     private val now = ngsiLdDateTime()
     private val attributeUuid = UUID.randomUUID()
     private val entityId = "urn:ngsi-ld:BeeHive:${UUID.randomUUID()}".toUri()
 
+    @BeforeEach
```
thomasBousselin, line 53: have you tried using @TestPropertySource(properties = [""])

bobeal: it will apply to all my tests and I need a special config for one test in the class

thomasBousselin (approved):

PR #1255, opened 2024-11-01
chore: activate more Detekt rules

bobeal:


search-service/src/test/kotlin/com/egm/stellio/search/entity/service/EntityServiceQueriesTests.kt
```diff
@@ -281,8 +297,8 @@ class EntityServiceQueriesTests : WithTimescaleContainer, WithKafkaContainer {
         "listOfString==\"iot\", 1, urn:ngsi-ld:BeeHive:01",
         "listOfString==\"data's processing\", 1, urn:ngsi-ld:BeeHive:01",
         "listOfString==\"stellio\", 2, 'urn:ngsi-ld:BeeHive:01,urn:ngsi-ld:BeeHive:02'",
```
ranim-n, line 299: why not use multiline string here too since `"` is used and needs to be escaped too?

search-service/src/main/kotlin/com/egm/stellio/search/entity/service/EntityService.kt
```diff
@@ -574,21 +599,20 @@ class EntityService(
...
+                entityId,
+                attributeName,
+                datasetId
+            ).bind()
+            entityAttributeService.deleteAttribute(
+                entityId,
+                attributeName,
+                datasetId,
+                deleteAll
+            ).bind()
```
thomasBousselin, line 615: I think the new implementation is less clear.
I am not sure there is a need to enforce this rule.

bobeal: actually a standard Kotlin convention: https://kotlinlang.org/docs/coding-conventions.html#if-versus-when

but I agree the previous version was a bit more readable

ranim-n (approved):

search-service/src/test/kotlin/com/egm/stellio/search/discovery/service/EntityTypeServiceTests.kt
```diff
@@ -15,9 +15,34 @@ import com.egm.stellio.search.support.WithTimescaleContainer
...
+import com.egm.stellio.shared.util.OUTGOING_COMPACT_PROPERTY
+import com.egm.stellio.shared.util.OUTGOING_PROPERTY
+import com.egm.stellio.shared.util.SENSOR_COMPACT_TYPE
+import com.egm.stellio.shared.util.SENSOR_TYPE
+import com.egm.stellio.shared.util.TEMPERATURE_PROPERTY
+import com.egm.stellio.shared.util.ngsiLdDateTime
+import com.egm.stellio.shared.util.shouldFail
+import com.egm.stellio.shared.util.shouldSucceedWith
+import com.egm.stellio.shared.util.toUri
+import com.egm.stellio.shared.util.typeNotFoundMessage
```
thomasBousselin, line 45: I think this number of import can be an issue.
But i think we should solve it by putting the properties/terms/types in objects instead of removing the rule.

thomasBousselin (approved):

search-service/src/main/kotlin/com/egm/stellio/search/entity/service/EntityService.kt
thomasBousselin: In the rest of the PR, the swap from 'when' to 'if' was an improvement. So it makes sense to keep the rule

PR #1256, opened 2024-11-02
feat: add support for many EntitySelector objects in Query datatype

bobeal:


search-service/src/main/kotlin/com/egm/stellio/search/temporal/model/TemporalEntitiesQuery.kt
```diff
@@ -1,31 +1,55 @@
...
+}
+
+data class TemporalEntitiesQueryFromPost(
+    val entitiesQueryFromPost: EntitiesQueryFromPost,
+    override val temporalQuery: TemporalQuery,
+    override val withTemporalValues: Boolean,
+    override val withAudit: Boolean,
+    override val withAggregatedValues: Boolean
+) : TemporalEntitiesQuery(temporalQuery, withTemporalValues, withAudit, withAggregatedValues) {
+    override fun getEntitiesQuery(): EntitiesQuery = entitiesQueryFromPost
```
thomasBousselin, line 54: I think you can avoid deleting :
```
val entitiesQuery: EntitiesQuery
```
from TemporalEntitiesQuery.

 Then, TemporalEntitiesQueryFromPost and TemporalEntitiesQueryFromGet could override it with their respective types.

This approach would eliminate the need to create and call getEntitiesQuery().

bobeal: indeed, better. done.

search-service/src/main/kotlin/com/egm/stellio/search/temporal/model/TemporalEntitiesQuery.kt
```diff
@@ -40,16 +39,12 @@ data class TemporalEntitiesQueryFromGet(
     override val withTemporalValues: Boolean,
     override val withAudit: Boolean,
     override val withAggregatedValues: Boolean
-) : TemporalEntitiesQuery(temporalQuery, withTemporalValues, withAudit, withAggregatedValues) {
-    override fun getEntitiesQuery(): EntitiesQuery = entitiesQueryFromGet
-}
+) : TemporalEntitiesQuery(entitiesQueryFromGet, temporalQuery, withTemporalValues, withAudit, withAggregatedValues)
 
 data class TemporalEntitiesQueryFromPost(
     val entitiesQueryFromPost: EntitiesQueryFromPost,
```
thomasBousselin, line 45: is it possible to make it work with override val?

thomasBousselin (approved):

PR #1259, opened 2024-11-05
Upgrade to 3.3.5

thomasBousselin:
The configuration may need further refinement. 
I did a quick review of the specifications to determine what should be permitted.
 However, there are several new rules that might cause issues by blocking certain characters:
- the backslash
- the encoded period

or less likely
- the encoded '%'
- the null
- the encoded lineBreaks/tabulations

We could also define a pattern of valid parameter names/value

### DIFF
--- a/api-gateway/src/main/kotlin/com/egm/stellio/apigateway/config/FirewallConfig.kt
+++ b/api-gateway/src/main/kotlin/com/egm/stellio/apigateway/config/FirewallConfig.kt
@@ -1,4 +1,4 @@
-package com.egm.stellio.shared.config
+package com.egm.stellio.apigateway.config
 
 import org.springframework.beans.BeansException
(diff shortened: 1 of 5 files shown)


thomasBousselin:
There is no test for get attribute information with the expanded id in the test suite so i commented the test to wait for the next spring-boot release.

bobeal:
meanwhile, moved the tests to https://forge.etsi.org/rep/cim/ngsi-ld-test-suite/-/merge_requests/156

bobeal:
instead of having commented (and forever forgotten) code, better use https://junit.org/junit5/docs/5.0.0-M2/api/org/junit/jupiter/api/Disabled.html. you can have set a reason:

```
    @Disabled("Re-enable this test with spring-security 6.3.5")
```

bobeal (approved):

PR #1264, opened 2024-11-15
feat(common): simplify internal error handling

thomasBousselin:


shared/src/main/kotlin/com/egm/stellio/shared/model/ApiExceptions.kt
```diff
@@ -2,24 +2,164 @@ package com.egm.stellio.shared.model
...
+                    it.type = this.type
+                }
+            )
+    }
+
+    fun toServerWebExchange(exchange: ServerWebExchange): Mono<Void> {
+        logger.info("Returning server web exchange error ${this.type} (${this.message})")
+        exchange.response.statusCode = this.status
+        exchange.response.headers[CONTENT_TYPE] = MediaType.APPLICATION_JSON_VALUE
+        val body = serializeObject(this)
```
bobeal, line 51: here, you are also serializing the inherited `Exception` class, which is not the expected thing.

btw, we should also get rid of this inherited class (in FP, errors should not be exceptions). not sure of the impacts, to be checked.

shared/src/main/kotlin/com/egm/stellio/shared/model/ApiExceptions.kt
```diff
@@ -2,24 +2,164 @@ package com.egm.stellio.shared.model
...
+}
+
+data class AlreadyExistsException(override val message: String) : APIException(
+    ErrorType.ALREADY_EXISTS.type,
+    HttpStatus.CONFLICT,
+    "The referred element already exists",
+    message
+)
+
+data class InvalidRequestException(override val message: String) : APIException( // todo check
```
bobeal, line 65: a bit surprising that it is marked as unused

shared/src/main/kotlin/com/egm/stellio/shared/model/ErrorResponse.kt
```diff
@@ -1,119 +0,0 @@
-package com.egm.stellio.shared.model
-
-import java.net.URI
-
-sealed class ErrorResponse(
```
bobeal, line 5: it should be the ultimate hierarchy for representing errors (https://old.arrow-kt.io/docs/patterns/error_handling/)

shared/src/main/kotlin/com/egm/stellio/shared/model/ApiExceptions.kt
```diff
@@ -2,24 +2,129 @@ package com.egm.stellio.shared.model
...
+)
+
+data class UnsupportedMediaTypeStatusApiException(override val message: String) : APIException(
+    ErrorType.UNSUPPORTED_MEDIA_TYPE.type,
+    HttpStatus.UNSUPPORTED_MEDIA_TYPE,
+    message
+)
+
+data class JsonLdErrorApiResponse(override val message: String, override val detail: String) : APIException(
+    ErrorType.BAD_REQUEST_DATA.type, // todo check
```
bobeal, line 123: yes: "If the data included by the JSON-LD document is not syntactically correct, according to the @context or the API data type definitions, then an error of type BadRequestData shall be raised"

shared/src/main/kotlin/com/egm/stellio/shared/model/ApiExceptions.kt
```diff
@@ -2,24 +2,129 @@ package com.egm.stellio.shared.model
...
+import org.springframework.http.HttpStatus
+import org.springframework.http.MediaType
+import org.springframework.http.ProblemDetail
+import org.springframework.http.ResponseEntity
+import java.net.URI
+
+const val DEFAULT_DETAIL = "If you have difficulty identifying the issue," +
+    "you can check common problem on https://stellio.readthedocs.io/en/latest/TROUBLESHOOT.html" +
+    "If you believe this is a bug, you can report it by creating an issue on" +
+    "https://github.com/stellio-hub/stellio-context-broker"
```
bobeal, line 17:
```suggestion
val DEFAULT_DETAIL =
    """
        If you have difficulty identifying the exact cause of the error,
        please check the list of some usual causes on https://stellio.readthedocs.io/en/latest/TROUBLESHOOT.html.
        If the error is still not clear or if you think it is a bug, feel free to open an issue on
        https://github.com/stellio-hub/stellio-context-broker
    """.trimIndent()
```

shared/src/main/kotlin/com/egm/stellio/shared/model/ApiExceptions.kt
```diff
@@ -2,24 +2,129 @@ package com.egm.stellio.shared.model
...
-data class NotImplementedException(override val message: String) : APIException(message)
-data class LdContextNotAvailableException(override val message: String) : APIException(message)
-data class NonexistentTenantException(override val message: String) : APIException(message)
-data class NotAcceptableException(override val message: String) : APIException(message)
+    val type: URI,
+    @JsonIgnore
+    val status: HttpStatus,
+    @JsonProperty("detail")
+    override val message: String,
+    open val detail: String = DEFAULT_DETAIL
```
bobeal, line 25: the value of the `JsonProperty` is kind of conflicting with the `detail` member.

bobeal: don't think the Jackson annotations are even used since it is always serialized as a `ProblemDetails` structure?

shared/src/main/kotlin/com/egm/stellio/shared/model/ApiExceptions.kt
thomasBousselin: Sadly the responseEntity mapper dislike line breaks. So this put "\n" in the return message.
```json
    "detail": "If you have difficulty identifying the exact cause of the error,\nplease check the list of some usual causes on https://stellio.readthedocs.io/en/latest/TROUBLESHOOT.html.\nIf the error is still not clear or if you think it is a bug, feel free to open an issue on\nhttps://github.com/stellio-hub/stellio-context-broker",
```
maybe it can be configured to work properly.

bobeal: don't bother with this, go back to the string concat version (but keep with the suggested content :) )

thomasBousselin: We can't remove them because the encoding of the line break come from the json specification : https://www.[ietf.org/rfc/rfc4627.txt](https://www.ietf.org/rfc/rfc4627.txt)

bobeal (approved):

ranim-n (approved):

PR #1274, opened 2024-11-28
feat: add language filter to sub attributes

ranim-n:


shared/src/main/kotlin/com/egm/stellio/shared/model/CompactedEntity.kt
```diff
@@ -140,27 +158,41 @@ private fun filterLanguageProperty(value: Map<String, Any>, transformationParame
...
+                propertyLocales.first()
+            }
+        val filteredMainAttribute = mutableMapOf(
+            JSONLD_TYPE_TERM to NGSILD_PROPERTY_TERM,
+            JSONLD_VALUE_TERM to (value[JSONLD_LANGUAGEMAP_TERM] as Map<String, Any>)[bestLocaleMatch],
+            NGSILD_LANG_TERM to bestLocaleMatch
+        )
+
+        value.filter { entry ->
+            !JSONLD_COMPACTED_ATTRIBUTE_CORE_MEMBERS.contains(entry.key) && entry.key != JSONLD_LANGUAGEMAP_TERM
```
thomasBousselin, line 177: you can replace entry
```suggestion
        value.filter { (key,_) ->
            !JSONLD_COMPACTED_ATTRIBUTE_CORE_MEMBERS.contains(key) && key != JSONLD_LANGUAGEMAP_TERM
```

shared/src/main/kotlin/com/egm/stellio/shared/model/CompactedEntity.kt
```diff
@@ -140,27 +158,41 @@ private fun filterLanguageProperty(value: Map<String, Any>, transformationParame
...
+        }.forEach { entry ->
```
thomasBousselin, line 178: same here you can use (key,value) instead of entry

shared/src/main/kotlin/com/egm/stellio/shared/model/CompactedEntity.kt
```diff
@@ -140,27 +158,41 @@ private fun filterLanguageProperty(value: Map<String, Any>, transformationParame
...
+            }
+        }
+        return filteredMainAttribute
+    } else return value.map { entry ->
+        when {
+            entry.key == NGSILD_ENTITY_TERM ->
+                entry.key to (entry.value as CompactedEntity).toFilteredLanguageProperties(languageFilter)
+            !JSONLD_COMPACTED_ATTRIBUTE_CORE_MEMBERS.contains(entry.key) ->
+                entry.key to filterLanguageProperty(entry.value as Map<String, Any>, transformationParameters)
+            else -> entry.key to entry.value
```
thomasBousselin, line 193: if you want to keep entry as the argument (as you want) you can use entry.toPair()

shared/src/main/kotlin/com/egm/stellio/shared/model/CompactedEntity.kt
```diff
@@ -42,6 +46,20 @@ typealias CompactedEntity = Map<String, Any>
 typealias CompactedAttributeInstance = Map<String, Any>
 typealias CompactedAttributeInstances = List<CompactedAttributeInstance>
 
+val JSONLD_COMPACTED_ATTRIBUTE_CORE_MEMBERS =
```
bobeal, line 49: `JSONLD_LANGUAGEMAP_TERM` should be in the list (if it is not in the list, it is a trap for future users of your list)

shared/src/main/kotlin/com/egm/stellio/shared/model/CompactedEntity.kt
bobeal: current is easier to read as it is similar to other branches in the when

bobeal:
I added a test that fails (but should not )

bobeal (approved):
did some changes to remove detekt warning and more fluent processing

thomasBousselin (approved):

PR #1281, opened 2024-12-08
feat: add support for `deletedAt` and NGSI-LD Null

bobeal:
First iteration with support in core and temporal API


search-service/src/main/kotlin/com/egm/stellio/search/entity/service/EntityQueryService.kt
```diff
@@ -251,15 +267,34 @@ class EntityQueryService(
...
     }
 
+    /**
+     * Used for checks before creating a (temporal) entity. Allows to know if the entity does not exist,
+     * or, if it exists, whether it is currently deleted (in which case, it may be possible to create it again
+     * if authorized)
+     */
+    suspend fun getEntityState(
+        entityId: URI
+    ): Either<APIException, Pair<URI, ZonedDateTime?>> {
```
thomasBousselin, line 284: Why are we returning the id? Is it different from the entityId argument? Could also use a better name maybe something like getEntityDeletionDate .

search-service/src/main/kotlin/com/egm/stellio/search/temporal/service/TemporalService.kt
```diff
@@ -64,14 +72,7 @@ class TemporalService(
...
-        // create a view of the entity containing only the most recent instance of each attribute
-        val expandedEntity = ExpandedEntity(
-            sortedJsonLdInstances
-                .keepFirstInstances()
-                .addCoreMembers(jsonLdTemporalEntity.id, jsonLdTemporalEntity.types)
-        )
-        val ngsiLdEntity = expandedEntity.toNgsiLdEntity().bind()
-
+        val (expandedEntity, ngsiLdEntity) = parseExpandedInstances(sortedJsonLdInstances, jsonLdTemporalEntity).bind()
         entityService.createEntity(ngsiLdEntity, expandedEntity, sub).bind()
```
thomasBousselin, line 76: Maybe you should have an already exist error here even if the entity is marked as deleted.

search-service/src/main/kotlin/com/egm/stellio/search/entity/service/EntityQueryService.kt
ranim-n: I don't think `getEntityDletionDate()` is fit because if the entity doesn't exist it returns a `ResourceNotFoundException` -> therefore allowing to know the state of the entity
Maybe explain more in the comment instead?

thomasBousselin: In the other hand the function is returning the deletedAt property so it is getting the deletionDate. I see that it is only used to know the actual deleted state of the entity so it could return a boolean (or enum) that represent if it was deleted. (in this case a possible name will be isMarkedDeleted)

search-service/src/main/kotlin/com/egm/stellio/search/entity/service/EntityAttributeService.kt
```diff
@@ -329,79 +298,122 @@ class EntityAttributeService(
...
+        val attributesToDeleteWithPayload = attributesToDelete
+            .map {
+                Triple(
+                    it,
+                    deletedAt,
+                    JsonLdUtils.expandAttribute(
+                        it.attributeName,
+                        it.attributeType.toNullCompactedRepresentation(),
+                        listOf(applicationProperties.contexts.core)
+                    ).second[0]
```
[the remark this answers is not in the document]
bobeal: indeed

search-service/src/main/kotlin/com/egm/stellio/search/entity/service/EntityQueryService.kt
bobeal: yep, I will change that. and yes, returning the id of the entity is useless.

search-service/src/main/kotlin/com/egm/stellio/search/entity/service/EntityAttributeService.kt
```diff
@@ -329,79 +298,122 @@ class EntityAttributeService(
...
-            AND attribute_name = :attribute_name
+            RETURNING id
             """.trimIndent()
         )
             .bind("entity_id", entityId)
-            .bind("attribute_name", attributeName)
-            .execute()
+            .allToMappedList { toUuid(it["id"]) }
+
+        attributeInstanceService.deleteInstancesOfEntity(deletedTeas).bind()
```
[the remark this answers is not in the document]
bobeal: ?

search-service/src/main/kotlin/com/egm/stellio/search/entity/util/AttributeUtils.kt
```diff
@@ -131,6 +136,23 @@ fun guessPropertyValueType(
         else -> Pair(Attribute.AttributeValueType.STRING, Triple(value.toString(), null, null))
     }
 
+/**
+ * Returns whether the expanded attribute instance holds a NGSI-LD Null value
+ */
+fun hasNgsiLdNullValue(
+    attribute: Attribute,
+    expandedAttributeInstance: ExpandedAttributeInstance
+): Boolean =
```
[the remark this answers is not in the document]
bobeal: swap parameters + use a Pair(attributeName, attributeType)

search-service/src/main/kotlin/com/egm/stellio/search/temporal/service/TemporalService.kt
bobeal: what is done in `EntityService` when you try to create an entity that already exists
- if not marked as deleted: already exists error (as before)
- if marked as deleted: ok if you are admin of the entity, forbidden if not

does not make sense?

search-service/src/main/kotlin/com/egm/stellio/search/entity/service/EntityAttributeService.kt
```diff
@@ -99,6 +104,12 @@ class EntityAttributeService(
             VALUES 
                 (:id, :entity_id, :attribute_name, :attribute_type, :attribute_value_type, :created_at, :dataset_id, 
                     :payload)
+            ON CONFLICT (entity_id, attribute_name, dataset_id)
```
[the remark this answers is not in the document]
bobeal: yes

search-service/src/main/kotlin/com/egm/stellio/search/temporal/service/TemporalService.kt
thomasBousselin: I thought that we were not supposed to create a temporal entity if it already existed. But that is not the case. The post endpoint is specified as update Or Create.

search-service/src/main/kotlin/com/egm/stellio/search/entity/service/EntityAttributeService.kt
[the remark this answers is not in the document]
thomasBousselin: oh, the second delete is for the attribute_instance table.
i got confused because we filter on a column named "temporal_entity_attribute". Can we easily rename her temporal_entity_attribute_id?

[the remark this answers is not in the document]
bobeal: we could, but not in the PR.

search-service/src/test/kotlin/com/egm/stellio/search/support/WithKafkaContainer.kt
```diff
@@ -1,29 +1,21 @@
...
+import org.springframework.boot.testcontainers.service.connection.ServiceConnection
+import org.testcontainers.junit.jupiter.Container
+import org.testcontainers.junit.jupiter.Testcontainers
 import org.testcontainers.kafka.ConfluentKafkaContainer
-import org.testcontainers.utility.DockerImageName
 
-interface WithKafkaContainer {
+@Testcontainers
+@Suppress("UtilityClassWithPublicConstructor")
+open class WithKafkaContainer {
```
thomasBousselin, line 10: What was the issue? 
Because in kotlin you can only inherit from one class at a time so this will limit what we can do with tests in the futur.

thomasBousselin (approved):

PR #1290, opened 2024-12-18
fix: incorrect behaviors in Batch Entity Upsert endpoint

bobeal:
- noOverwrite option is not handled whereas it should be passed when in update mode
- replace mode was only replacing the attributes of the entity (and thus was missing all the other needed actions: replace types and scopes, send events...)

### DIFF
--- a/search-service/src/main/kotlin/com/egm/stellio/search/entity/web/EntityOperationHandler.kt
+++ b/search-service/src/main/kotlin/com/egm/stellio/search/entity/web/EntityOperationHandler.kt
@@ -109,7 +109,9 @@ class EntityOperationHandler(
         )
         @RequestParam queryParams: MultiValueMap<String, String>
     ): ResponseEntity<*> = either {
-        val options = queryParams.getFirst(QP.OPTIONS.key)
+        val options = queryParams.getFirst(QP.OPTIONS.key)?.split(",")
+        val disallowOverwrite = options?.any { it == OptionsValue.NO_OVERWRITE.value } == true
+        val updateMode = options?.any { it == OptionsValue.UPDATE_MODE.value } == true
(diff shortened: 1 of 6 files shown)


search-service/src/main/kotlin/com/egm/stellio/search/entity/web/EntityOperationHandler.kt
```diff
@@ -109,7 +109,9 @@ class EntityOperationHandler(
         )
         @RequestParam queryParams: MultiValueMap<String, String>
     ): ResponseEntity<*> = either {
-        val options = queryParams.getFirst(QP.OPTIONS.key)
+        val options = queryParams.getFirst(QP.OPTIONS.key)?.split(",")
+        val disallowOverwrite = options?.any { it == OptionsValue.NO_OVERWRITE.value } == true
```
thomasBousselin, line 113: I don't see a no_overwrite option for upsert in the specification.

bobeal: see the in-progress doc with misc fixes to contribute to the spec

ranim-n (approved):

search-service/src/main/kotlin/com/egm/stellio/search/entity/web/EntityOperationHandler.kt
bobeal: as discussed, aligned with the figures of operations from 1.6.1. Spec needs clarification, to be discussed with ISG-CIM.

bobeal:
@thomasBousselin OK for you?

thomasBousselin (approved):

PR #1319, opened 2025-01-18
feat: add support for user defined geo properties

bobeal:


thomasBousselin:
It will be easier to review if it was based on the "chore/remove-unused-observation-listener" branch.

bobeal:
> It will be easier to review if it was based on the "chore/remove-unused-observation-listener" branch.

indeed, forgot to do it.

bobeal:
> > It will be easier to review if it was based on the "chore/remove-unused-observation-listener" branch.
> 
> indeed, forgot to do it.

and done.

ranim-n (approved):

thomasBousselin (commented):
So the "geometryProperty" parameter was only working for location, observationSpace and operationSpace?
 Have you checked that it work with the other properties?

bobeal:
> So the "geometryProperty" parameter was only working for location, observationSpace and operationSpace? Have you checked that it work with the other properties?

The main point is to fix the issue referenced.

As for  the `geometryProperty`, it is another thing (and the code is dumb on this, just trying to get the property with the name passed as parameter)

thomasBousselin (approved):

PR #1327, opened 2025-02-03
fix(temp): use deleted entity in notifications

bobeal:
Temporary solution while `showChanges` is not implemented. It allows to have access to attributes in case of a `entityDeleted` event. For other events, previous and updated entity are the same.


ranim-n (approved):

thomasBousselin:
> For other events, previous and updated entity are the same.
Strange behavior. but if you double checked this its ok.

Also maybe add comment.

bobeal:
> > For other events, previous and updated entity are the same.
> 
> Strange behavior.

Temporary while `showChanges` is not implemented. At the end, the previous entities will be conveyed with the events and will be used to show changes when asked.

thomasBousselin (approved):

PR #1350, opened 2025-03-05
Feature/modifiedAt at creation

thomasBousselin:


search-service/src/main/kotlin/com/egm/stellio/search/entity/model/Attribute.kt
```diff
@@ -45,7 +45,7 @@ data class Attribute(
     val attributeValueType: AttributeValueType,
     val datasetId: URI? = null,
     val createdAt: ZonedDateTime,
-    val modifiedAt: ZonedDateTime? = null,
+    val modifiedAt: ZonedDateTime? = createdAt, // todo could be non nullable
```
bobeal, line 48: this TODO has to be removed. and, yes, since we now automatically set a `modifiedAt` when creating an entity or attribute, it should be non nullable.

bobeal: and what about https://github.com/stellio-hub/stellio-context-broker/blob/develop/search-service/src/main/kotlin/com/egm/stellio/search/entity/model/Entity.kt#L25? (it should also be set for entity, not only attributes)

search-service/src/main/kotlin/com/egm/stellio/search/entity/web/EntityHandler.kt
```diff
@@ -221,9 +221,9 @@ class EntityHandler(
             implemented = [
```
bobeal, line 221: modifications in this file should not be in this PR

shared/src/main/kotlin/com/egm/stellio/shared/model/ExpandedEntity.kt
```diff
@@ -53,12 +53,13 @@ data class ExpandedEntity(
                 if (JSONLD_EXPANDED_ENTITY_CORE_MEMBERS.contains(it.key))
                     it.value
                 else castAttributeValue(it.value).map { expandedAttributeInstance ->
-                    expandedAttributeInstance.addDateTimeProperty(
-                        NGSILD_CREATED_AT_PROPERTY,
-                        createdAt
-                    ) as ExpandedAttributeInstance
+                    expandedAttributeInstance
```
bobeal, line 56: as the name of the function says, it is called at entity creation time, but you are missing attributes creation times (and entity replace?).

shared/src/main/kotlin/com/egm/stellio/shared/model/ExpandedEntity.kt
```diff
@@ -94,7 +95,7 @@ data class ExpandedEntity(
             as List<ExpandedTerm>
     }
 
-    fun toTypeSelection() = types.joinToString { "|" }
+    fun toTypeSelection() = types.joinToString("|")
```
bobeal, line 98: lost modification

bobeal:
take care of the name of the PR. at the end, the squashed commit should follow https://github.com/stellio-hub/stellio-context-broker?tab=readme-ov-file#committing

search-service/src/main/kotlin/com/egm/stellio/search/entity/service/EntityAttributeService.kt
```diff
@@ -220,7 +220,7 @@ class EntityAttributeService(
 
         // if the temporal property existed before, the create operation returned a different id than the one
         // in the attribute object
-        val timeProperty =
+        val timeProperty = // todo what is the point of choosing this?
```
bobeal, line 223: for the audit trace added below, we need to know if it was a creation or a modification (the `create(Attribute)` is actually an upsert)

search-service/src/main/kotlin/com/egm/stellio/search/temporal/model/AttributeInstance.kt
```diff
@@ -67,6 +67,7 @@ data class AttributeInstance private constructor(
             sub = sub
         )
 
+        // todo can i delete this function ?
```
bobeal, line 70: if we don't loose the `instanceId` thing

bobeal (approved):

bobeal:
can be merged?

PR #1355, opened 2025-03-08
chore: clean and automatically check unnecessary stubs

bobeal:
- upgrade to last version of mockk.io (since springmockk is not updated for a long time)

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -82,6 +82,7 @@ subprojects {
         testImplementation("org.springframework.boot:spring-boot-testcontainers")
         testImplementation("io.projectreactor:reactor-test")
         testImplementation("com.ninja-squad:springmockk:4.0.2")
+        testImplementation("io.mockk:mockk:1.13.17")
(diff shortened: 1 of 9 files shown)


thomasBousselin:
So if i understand correctly, this check that all mock are used.
It found some in EntityOperationHandlerTest. 
And basically all the rest are false positive because the check does not understand wasNot Called?

bobeal:
> So if i understand correctly, this check that all mock are used. It found some in EntityOperationHandlerTest. And basically all the rest are false positive because the check does not understand wasNot Called?

The `wasNot` problem appeared with the upgrade of mockk.io (https://github.com/mockk/mockk/issues/1155). We have this dependency through spring-mockk but spring-mockk has not been updated for a while and thus we had a very outdated version of mockk.io.

bobeal:
> So if i understand correctly, this check that all mock are used. It found some in EntityOperationHandlerTest. And basically all the rest are false positive because the check does not understand wasNot Called?

And yes for the first point.

thomasBousselin (approved):

PR #1365, opened 2025-03-23
fix(temporal): allow retrieving temporal evolution of an entity when deleted from the current state

bobeal:
- manually validated, just missing some TCs


ranim-n (approved):

thomasBousselin:
Jenkins failed. The rest is ok.

bobeal:
> Jenkins failed. The rest is ok.

Two jobs at the same time 🙁

thomasBousselin (approved):

PR #1371, opened 2025-03-26
feat: align API operations with notification triggers

bobeal:


search-service/src/main/kotlin/com/egm/stellio/search/entity/model/Entity.kt
```diff
@@ -58,4 +59,7 @@ data class Entity(
                 NGSILD_DELETED_AT_PROPERTY to buildNonReifiedTemporalValue(deletedAt),
             )
         )
+
+    fun toExpandedEntity(): ExpandedEntity =
+        ExpandedEntity(this.payload.deserializeExpandedPayload())
```
thomasBousselin, line 64: So this only work if the entity is already expanded?

bobeal: yes, but it is always expanded in DB

search-service/src/main/kotlin/com/egm/stellio/search/entity/model/UpdateResult.kt
```diff
@@ -70,13 +70,12 @@ enum class OperationStatus {
...
     UPDATED,
     DELETED,
-    IGNORED,
     FAILED;
 
     fun isSuccessResult(): Boolean = getSuccessStatuses().contains(this)
 
     companion object {
-        fun getSuccessStatuses(): List<OperationStatus> = listOf(CREATED, UPDATED, DELETED, IGNORED)
+        fun getSuccessStatuses(): List<OperationStatus> = listOf(CREATED, UPDATED, DELETED)
```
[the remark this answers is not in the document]
bobeal: it was a personal, and I think wrong, interpretation of such a sentence in the specification: "If overwrite is not allowed, the existing Attribute with the specified datasetId in the target Entity shall be left untouched.". but it is better to consider it as a failed operation (something the user has asked for has not been done).

ranim-n (approved):

PR #1380, opened 2025-04-17
fix: early payload validation and incorrect payload update

bobeal:
- validate attributes payload before any DB operation
- fix a bug when updating an entity payload to modify the types

### DIFF
--- a/search-service/src/main/kotlin/com/egm/stellio/search/entity/service/EntityService.kt
+++ b/search-service/src/main/kotlin/com/egm/stellio/search/entity/service/EntityService.kt
@@ -8,6 +8,7 @@ import arrow.core.raise.either
 import arrow.core.right
 import arrow.core.toOption
 import com.egm.stellio.search.authorization.service.AuthorizationService
+import com.egm.stellio.search.common.util.deserializeAsMap
 import com.egm.stellio.search.common.util.deserializeExpandedPayload
 import com.egm.stellio.search.common.util.execute
 import com.egm.stellio.search.common.util.oneToResult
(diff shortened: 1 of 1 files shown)


search-service/src/main/kotlin/com/egm/stellio/search/entity/service/EntityService.kt
```diff
@@ -206,6 +208,12 @@ class EntityService(
                 currentAttributeName == newAttributeName && currentDatasetId == newDatasetId
             }
         }
+        val attributesToCreateOrUpdate = newEntityAttributes
+            .map { (attributeName, _, expandedAttributeInstance) ->
+                val expandedAttributes = mapOf(attributeName to listOf(expandedAttributeInstance))
+                val ngsiLdAttributes = expandedAttributes.toNgsiLdAttributes().bind()
+                Pair(expandedAttributes, ngsiLdAttributes)
```
thomasBousselin, line 215: kotlin notation : `expandedAttributes to ngsiLdAttributes`

thomasBousselin (commented):
This should be taken care of by the @Transactionnal The best way to fix this would be to make it trigger when returning a Either.left. (I don't know if it is possible)

search-service/src/main/kotlin/com/egm/stellio/search/entity/service/EntityService.kt
```diff
@@ -161,12 +162,13 @@ class EntityService(
                 .filter { it.first != JSONLD_ID }
                 .partition { JSONLD_EXPANDED_ENTITY_SPECIFIC_MEMBERS.contains(it.first) }
         val mergedAt = ngsiLdDateTime()
+        val ngsiLdAttributes = otherAttrs.toMap().toNgsiLdAttributes().bind()
```
ranim-n, line 165: why did you add this variable?

bobeal:
> This should be taken care of by the @Transactionnal The best way to fix this would be to make it trigger when returning a Either.left. (I don't know if it is possible)

And most of the time, this is not what we want. For instance, in the sample described in the issue, it should return a 207 saying:
- success for `type` and `isEnabled`
- error for `datasetId` and `object`

Now that I think about it, I should probably restrict the PR to just fix the update of the entity payload when updating the types. The atomicity of attributes updates will be a bigger refactoring to handle in another PR...

search-service/src/main/kotlin/com/egm/stellio/search/entity/service/EntityService.kt
bobeal: to have an early validation of the received payload. maybe not so a great idea: https://github.com/stellio-hub/stellio-context-broker/pull/1380#issuecomment-2812174466

PR #1404, opened 2025-05-29
feat(subscription): add `triggerReason` in notifications

bobeal:


ranim-n (approved):

subscription-service/src/main/kotlin/com/egm/stellio/subscription/job/TimeIntervalNotificationJob.kt
```diff
@@ -66,6 +67,8 @@ class TimeIntervalNotificationJob(
     ): Triple<Subscription, Notification, Boolean> =
         notificationService.callSubscriber(
             subscription,
+            // recurring subscriptions do not really apply to notification triggers, take the 1st as a default
```
thomasBousselin, line 70: It do not apply at all(since timeInterval and watchedAttributes are mutually exclusive).
So it is strange to send a random reason.
 Maybe  add a new NotificationTrigger reason for periodic Subscriptions.

thomasBousselin (commented):
I would still prefer to have an update from the specification on this.

PR #1414, opened 2025-06-23
refactor: use DataTypes for subscriptions

bobeal:
- fix a bug in DataTypes.toFinalRepresentation where first let was actually not applied
- more SQL related functions in ParsingUtils to DBConversionUtils
- remove direct uses of DataType.mapper (internal object)
- remove some useless functions to serialize / deserialize in ParsingUtils


ranim-n (approved):

thomasBousselin (approved):
Permission are also a DataTypess o i will have to update my PR accordingly.

Very nice. We could go even further with this. By creating a DataTypes interface that provide the serialize out of the box. But need a compact function. (maybe later)

PR #1428, opened 2025-07-19
feat: extract dependencies into a specific docker compose file

bobeal:
- mainly to easily launch dependencies only while in development (and launch services in the IDE), without having to comment out the services in the docker compose file


thomasBousselin:
I use `docker compose up postgres kafka`. Wich work fine.

PR #1430, opened 2025-07-21
feat: check existence of assignee and target id

thomasBousselin:


search-service/src/main/kotlin/com/egm/stellio/search/authorization/permission/web/PermissionHandler.kt
```diff
@@ -225,7 +220,7 @@ class PermissionHandler(
 
         val permission = permissionService.getById(permissionId).bind()
 
-        if (permission.assignee !in subjects && checkIsAdmin(permissionId).isLeft()) {
+        if (permission.assignee !in subjects && checkCanModify(permissionId).isLeft()) {
```
bobeal, line 223: a bit counter intuitive to call `checkCanModify` when you want to read (though it was worse before)

thomasBousselin: Yes, but the twist is that you have to be admin of the target entity to read a permission. I think `checkIsAdmin(id)` was better.

search-service/src/main/kotlin/com/egm/stellio/search/authorization/permission/web/PermissionHandler.kt
```diff
@@ -326,16 +318,26 @@ class PermissionHandler(
...
-    private suspend fun checkIsAdmin(permission: Permission): Either<APIException, Unit> =
-        permissionService.checkHasPermissionOnEntity(permission.target.id, Action.ADMIN)
-            .flatMap {
-                if (!it)
-                    AccessDeniedException(
-                        unauthorizedCreateMessage(permission.target.id)
-                    ).left()
-                else
-                    Unit.right()
+    private suspend fun checkCanModify(
```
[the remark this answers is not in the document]
thomasBousselin: Yes, i shouldn't have rename the function that work with the id.

search-service/src/main/kotlin/com/egm/stellio/search/authorization/permission/web/PermissionHandler.kt
```diff
@@ -326,16 +318,26 @@ class PermissionHandler(
...
+        permission: Permission,
+        isDelete: Boolean = false
+    ): Either<APIException, Unit> = either {
+        if (permission.action == Action.OWN) {
+            CHANGE_OWNER_EXCEPTION.left().bind<APIException>()
+        }
```
bobeal, line 327: This check is also sometimes done in the handler function (for instance, in `update`). And indeed it is more a business check than an access right one and should not really be here.

search-service/src/main/kotlin/com/egm/stellio/search/authorization/permission/web/PermissionHandler.kt
```diff
@@ -326,16 +318,26 @@ class PermissionHandler(
...
+
+        val hasPermission = permissionService.checkHasPermissionOnEntity(permission.target.id, Action.ADMIN).bind()
+
+        if (!hasPermission)
+            AccessDeniedException(unauthorizedCreateMessage(permission.target.id)).left().bind()
```
bobeal, line 332: Error message will not always be appropriate since you call it from `create` and `delete`

search-service/src/main/kotlin/com/egm/stellio/search/authorization/permission/web/PermissionHandler.kt
bobeal, line 321: Having a `checkCanCreate` and `checkCanDelete` will make things a lot more clear (and will actually duplicate only two lines of code)

PR #1449, opened 2025-08-23
refactor(core): avoid retrieving entity from DB in events

bobeal:
- instead use the updated entity payload as just persisted in DB in EntityService


search-service/src/main/kotlin/com/egm/stellio/search/entity/service/EntityEventService.kt
```diff
@@ -270,24 +260,4 @@ class EntityEventService(
...
-        this.fold({
-            logger.error(
-                "Error sending {} event for entity {} in tenant {}: {}",
-                eventCategory,
-                entityId,
-                tenantName,
-                it
-            )
-        }, {
-            logger.debug("Sent {} event for entity {} in tenant {}", eventCategory, entityId, tenantName)
```
thomasBousselin, line 291: What motivate this removal?

thomasBousselin (approved):

ranim-n (dismissed):

search-service/src/main/kotlin/com/egm/stellio/search/entity/service/EntityEventService.kt
bobeal: the error log was actually called when retrieving the entity failed (and not when sending to Kafka failed...), so it is now useless.

the debug log is mostly redundant with already existing logs.

PR #1453, opened 2025-08-26
fix: fix migration when duplicated global permission

thomasBousselin:


bobeal (approved):

bobeal:
@thomasBousselin to close?

PR #1471, opened 2025-09-25
chore: add reference to Conventional Branch specification

bobeal:


thomasBousselin (dismissed):
So what name should have a branch used for doing refactoring?

bobeal:
> So what name should have a branch used for doing refactoring?

Not clear (https://conventional-branch.github.io/#why-arent-branch-types-as-detailed-as-conventional-commits-eg-build-ci-docs-style-refactor). But we can agree to still use the `refactor/` prefix.

bobeal:
> > So what name should have a branch used for doing refactoring?
> 
> Not clear (https://conventional-branch.github.io/#why-arent-branch-types-as-detailed-as-conventional-commits-eg-build-ci-docs-style-refactor). But we can agree to still use the `refactor/` prefix.

btw, I'll update the doc to list authorized prefixes.

PR #1498, opened 2025-10-24
fix: ignored allow-public-permission parameter

thomasBousselin:


shared/src/main/kotlin/com/egm/stellio/shared/config/ApplicationProperties.kt
```diff
@@ -10,7 +10,8 @@ data class ApplicationProperties(
     val contexts: Contexts
 ) {
     data class Authentication(
-        val enabled: Boolean
+        val enabled: Boolean,
+        val allowPublicPermission: Boolean
```
bobeal: don't see why you would need that since you don't access the property from this object.

PR #1509, opened 2025-11-06
chore: configure WebFlux using Spring configuration properties

bobeal:
- see https://docs.spring.io/spring-boot/appendix/application-properties/index.html#application-properties.web.spring.http.codecs.log-request-details
- see https://docs.spring.io/spring-boot/appendix/application-properties/index.html#application-properties.web.spring.http.codecs.max-in-memory-size

### DIFF
--- a/shared/src/main/resources/shared.properties
+++ b/shared/src/main/resources/shared.properties
@@ -28,7 +28,6 @@ spring.flyway.baseline-on-migrate=true
 spring.flyway.baseline-version=0
 spring.flyway.enabled=false
 
-spring.http.codecs.log-request-details=true
 spring.http.codecs.max-in-memory-size=2097152
 
 server.error.include-stacktrace = always


shared/src/main/resources/shared.properties
```diff
@@ -28,6 +28,9 @@ spring.flyway.baseline-on-migrate=true
 spring.flyway.baseline-version=0
 spring.flyway.enabled=false
 
+spring.http.codecs.log-request-details=true
```
thomasBousselin: true par default?

bobeal: it is false by default (see the link in the description)

thomasBousselin: i meant for us, do we want to put it at true by default?

thomasBousselin (dismissed):

PR #1511, opened 2025-11-12
fix: prevent error when requesting a deleted assignee

thomasBousselin:


search-service/src/main/kotlin/com/egm/stellio/search/authorization/permission/service/PermissionService.kt
```diff
@@ -489,10 +489,7 @@ class PermissionService(
         }
 
         val assigneeFilter = permissionFilters.assignee?.let {
-            val assignee = permissionFilters.assignee
-
-            val assigneeUuids = subjectReferentialService.getSubjectAndGroupsUUID(assignee).bind()
-            buildIsAssigneeFilter(assigneeUuids)
+            buildIsAssigneeFilter(listOf(it))
```
bobeal: we are "losing" the permissions granted from groups memberships? this is worrying since groups memberships are highly used.

PR #1538, opened 2025-12-01
feat: add support for Delete and Reload @context endpoint

bobeal:


bobeal:
Not totally ready (I miss the reload and maybe some minor improvements), but ready for a first review.

PR #1549, opened 2025-12-05
feat(common): add partial implementation of Entity Ordering

thomasBousselin:


search-service/src/main/kotlin/com/egm/stellio/search/common/model/OrderBy.kt
```diff
@@ -0,0 +1,41 @@
...
+import arrow.core.raise.either
+import com.egm.stellio.shared.model.APIException
+import com.egm.stellio.shared.model.InvalidRequestException
+import com.egm.stellio.shared.queryparameter.AttributePath
+
+data class OrderBy(
+    val param: String,
+    val contexts: List<String>
+) {
+    val direction: Direction = Direction.fromString(param.substringAfter(';')).fold({ Direction.ASC }, { it })
```
thomasBousselin: There is no validation of the OrderBy parameter.
I could add one, But kotlin arrow error handling don't work well with constructor.
- I could throw an error here directly. 
- Or I can do a companion function fromParam() who return an either
Both would return the same thing to the user.

search-service/src/main/kotlin/com/egm/stellio/search/common/model/OrderingParams.kt
```diff
@@ -0,0 +1,19 @@
+package com.egm.stellio.search.common.model
+
+/**
+ * A Query data type as defined in 5.2.23.
```
bobeal:
```suggestion
 * An OrderingParams data type as defined in 5.2.43.
```

shared/src/main/kotlin/com/egm/stellio/shared/queryparameter/AttributePath.kt
```diff
@@ -0,0 +1,65 @@
...
+
+            else ->
+                """$."${mainPath[0]}"."$NGSILD_RELATIONSHIP_OBJECT".$trailingPathString.**{0 to 1}."$JSONLD_ID_KW""""
+        }
+    }
+
+    fun buildSqlOrderClause() = """
+            jsonb_path_query_array(
+            entity_payload.payload,
+            '${buildJsonBPropertyPath()}'
```
bobeal: does that mean we cannot order on the value of a relationship?

search-service/src/main/kotlin/com/egm/stellio/search/entity/service/EntityQueryService.kt
```diff
@@ -76,19 +77,21 @@ class EntityQueryService(
...
-            SELECT DISTINCT(entity_payload.entity_id)
+            SELECT entity_payload.entity_id
             FROM entity_payload
             LEFT JOIN temporal_entity_attribute tea
             ON tea.entity_id = entity_payload.entity_id 
                 ${if (excludeDeleted) " AND tea.deleted_at is null " else ""}
             WHERE $filterQuery
             ${if (excludeDeleted) " AND entity_payload.deleted_at is null " else ""}
-            ORDER BY entity_id
+            GROUP BY entity_payload.entity_id, entity_payload.payload
```
bobeal: why this grouping?

search-service/src/main/kotlin/com/egm/stellio/search/temporal/service/TemporalQueryService.kt
```diff
@@ -171,7 +171,7 @@ class TemporalQueryService(
                 .toList()
                 // the ordering made when searching matching attributes is lost
                 // since we are now iterating over the map of attributes with their instances
-                .sortedBy { it.first.entityId }
+                .sortedBy { desiredOrder[it.first.entityId] }
```
bobeal: can't this work with `entitiesIds.indexOf(it.first.entityId)`?

search-service/src/test/kotlin/com/egm/stellio/search/entity/service/EntityServiceQueriesTests.kt
```diff
@@ -491,6 +493,43 @@ class EntityServiceQueriesTests : WithTimescaleContainer, WithKafkaContainer() {
         assertThat(entitiesIds).contains(entity02Uri)
     }
 
+    // possible ids :
+    // urn:ngsi-ld:BeeHive:01,urn:ngsi-ld:BeeHive:02
+    // urn:ngsi-ld:MultiTypes:03,urn:ngsi-ld:Beekeeper:04,
+    @ParameterizedTest
+    @CsvSource(
```
bobeal: miss a test on a relationship

thomasBousselin: How can you know that the attribute you are filtering on is a relationship? For the query you check if the value you filter is a URI. But Here i can't really do that.

I feel like the only solution is to always have two filters, one for the Property and one for Relationship. It is strange but it should work.

search-service/src/main/kotlin/com/egm/stellio/search/temporal/service/TemporalQueryService.kt
thomasBousselin: desiredOrder will be more performant if there is a lot of ids. But this is never the case. I can change it if you want.

shared/src/main/kotlin/com/egm/stellio/shared/queryparameter/AttributePath.kt
```diff
@@ -0,0 +1,63 @@
...
+            trailingPath.isEmpty() ->
+                """$."${mainPath[0]}"."$NGSILD_PROPERTY_VALUE"."$JSONLD_VALUE_KW""""
+            else ->
+                """$."${mainPath[0]}"."$NGSILD_PROPERTY_VALUE".$trailingPathString.**{0 to 1}."$JSONLD_VALUE_KW""""
+        }
+    }
+
+    fun buildJsonBRelationshipPath(): String {
+        val mainPathString = mainPath.joinToString(".") { "\"$it\"" }
+        val trailingPathString = mainPath.joinToString(".") { "\"$it\"" }
```
bobeal: it should be `trailingPath`, not `mainPath`

search-service/src/main/kotlin/com/egm/stellio/search/common/model/OrderBy.kt
```diff
@@ -0,0 +1,50 @@
...
+        }
+        return "$attributeSql $directionSql"
+    }
+    companion object {
+        fun fromParam(
+            param: String,
+            contexts: List<String>
+        ): Either<APIException, OrderBy> = either {
+            OrderBy(
+                Direction.fromString(param.substringAfter(';', "asc")).bind(),
```
bobeal: you could use `Direction.ASC.value`

search-service/src/main/kotlin/com/egm/stellio/search/common/model/OrderBy.kt
```diff
@@ -0,0 +1,50 @@
...
+    enum class Direction(val value: String) {
+        ASC("asc"),
+        DESC("desc"),
+        DIST_ASC("dist-asc"),
+        DIST_DESC("dist-desc");
+
+        companion object {
+            fun fromString(key: String): Either<APIException, Direction> = either {
+                Direction.entries.find { it.value == key }
+                    ?: return InvalidRequestException("'$key' is not a valid ordering direction parameter").left()
```
bobeal: it typically is a `BadRequestDataException`

search-service/src/test/kotlin/com/egm/stellio/search/entity/service/EntityServiceQueriesTests.kt
thomasBousselin: > how does it behave for multi-instances properties?

It compare the first instance with the first instance if they are equal it compare the second ...

search-service/src/main/kotlin/com/egm/stellio/search/temporal/service/TemporalQueryService.kt
bobeal: yes, does not seem really necessary

bobeal:
> And change the PR name to a compliant one.

Almost. Remember the PR name will become the commit message after the squash merge. So a compliant name would be: feat(common): add partial implementation of Entity Ordering

PR #1568, opened 2025-12-15
fix: rewrite content

bobeal:


thomasBousselin:
I added a commit with my fix.

PR #1576, opened 2025-12-26
feat(subscription): add support for cooldown and timeout

bobeal:


subscription-service/src/main/kotlin/com/egm/stellio/subscription/model/Subscription.kt
```diff
@@ -213,6 +215,16 @@ data class Subscription(
         else
             Unit.right()
 
+    private fun checkCooldownGreaterThanZero(): Either<APIException, Unit> =
```
thomasBousselin: checkCooldowIsValid?

bobeal: it is aligned with current naming in this file (`checkTimeIntervalGreaterThanZero`, `checkThrottlingGreaterThanZero`,...)

PR #1577, opened 2025-12-29
feat(core) add support for pick and omit in Linked Entities

bobeal:


shared/src/main/kotlin/com/egm/stellio/shared/model/AttributeProjection.kt
```diff
@@ -0,0 +1,232 @@
...
+ *           listOf(
+ *               AttributeProjection("temperature" to null),
+ *               AttributeProjection("observedAt", listOf(AttributeProjection("location" to null)))
+ *           )
+ *       )
+ * ```
+ */
+data class AttributeProjection(
+    val attributeName: String,
+    val nestedProjections: List<AttributeProjection>? = null
```
thomasBousselin: I think this serve the same purpose as the AttributePath from the OrderBy PR.
 Since the `relation{attribute}` notation also exist on query.

It is not critical since the q= does not support `relation{attribute}` yet. We can wait for the implementation  of `relation{attribute}` on q or we can merge the logic now.

shared/src/main/kotlin/com/egm/stellio/shared/model/AttributeProjection.kt
```diff
@@ -0,0 +1,232 @@
...
+        fun parsePickOmitParameters(
+            pickParam: String?,
+            omitParam: String?
+        ): Either<APIException, Pair<List<AttributeProjection>, List<AttributeProjection>>> = either {
+            val pick = pickParam?.let { pickValue ->
+                parseAttributeProjectionExpression(pickValue.replace(" ", ""))
+                    .mapLeft {
+                        if (it is BadRequestDataException)
+                            it.copy(message = "Error parsing pick parameter: ${it.message}")
+                        else it
```
thomasBousselin: To me, this logic is unnecessary. It would be better if we could return the error directly. Especially because there is no value in specifying wheter the error is in pick or omit, as they are not meant to be used together.

shared/src/main/kotlin/com/egm/stellio/shared/model/AttributeProjection.kt
```diff
@@ -0,0 +1,232 @@
...
+                    }
+                    .bind()
+            } ?: emptyList()
+
+            val omit = omitParam?.let { omitValue ->
+                parseAttributeProjectionExpression(omitValue.replace(" ", ""))
+                    .mapLeft {
```
thomasBousselin: same

shared/src/main/kotlin/com/egm/stellio/shared/model/AttributeProjection.kt
```diff
@@ -0,0 +1,232 @@
...
+ *       )
+ * ```
+ */
+data class AttributeProjection(
+    val attributeName: String,
+    val nestedProjections: List<AttributeProjection>? = null
+) {
+    companion object {
+
+        private val separators = setOf(',', '|')
```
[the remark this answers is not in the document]
bobeal: - Added the 1st point to my list of things to discuss in API call
- In the `Query` data type, it is meant to be expressed like `["temperature", "observedBy{title,name}", ...]` (but not sure to understand your point)

shared/src/main/kotlin/com/egm/stellio/shared/util/ListUtils.kt
thomasBousselin: Empty file

shared/src/main/kotlin/com/egm/stellio/shared/model/AttributeProjection.kt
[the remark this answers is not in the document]
bobeal: It should not. By chance, it will work. But not moving on this until fully clarified in the spec.

shared/src/main/kotlin/com/egm/stellio/shared/model/AttributeProjection.kt
```diff
@@ -0,0 +1,246 @@
...
+fun List<AttributeProjection>.removeAttributes(attributes: Set<String>): List<AttributeProjection> =
+    filter { it.attributeName !in attributes }
+
+fun List<AttributeProjection>.getAttributesToPickFor(parentAttributeName: String, level: UInt): Set<String> =
+    getAttributesFor(parentAttributeName, level, ProjectionType.PICK)
+
+fun List<AttributeProjection>.getAttributesToOmitFor(parentAttributeName: String, level: UInt): Set<String> =
+    getAttributesFor(parentAttributeName, level, ProjectionType.OMIT)
+
+private fun List<AttributeProjection>.getAttributesFor(
```
thomasBousselin: Will it work if you have multiple of the same name?
pick=X1{Y1{Z1}},X2{Y1{Z2}}}

thomasBousselin: No, to me it's a different one. One is having the same entity id on two different relationships. And the other is having the same relationships names on two subRelationShips.

bobeal: Giving up on this one, it adds too much complexity: it requires keeping track of the path from root to the current attribute for each attribute of each entity in the tree (globally a nightmare, not even speaking about the flat representation for which it needs a full rewrite)

bobeal: https://github.com/stellio-hub/stellio-context-broker/issues/1593

PR #1642, opened 2026-03-03
fix(docs): Add KC_DB=postgres to Keycloak integration compose example

0x69d:
Current example omits KC_DB.
Keycloak defaults to dev-file DB when db vendor not specified.
In some environments this results in Keycloak not persisting data to external PostgreSQL until KC_DB=postgres is set.
This PR adds KC_DB=postgres and a short note to prevent misconfiguration.

### DIFF
--- a/docs/admin/keycloak_integration.md
+++ b/docs/admin/keycloak_integration.md
@@ -54,6 +54,8 @@ The provided Docker image extends the official Keycloak Docker image to bundle i
 
 To start with, you can use this sample Docker compose file (do not forget to create a `.env` file with the environment variables used in the docker compose file):
 
+**Note:** By default, Keycloak uses a local `dev-file` database. Even when running with the `start-dev` command, it is highly recommended to set `KC_DB=postgres` as shown below to ensure your data is properly persisted to the external PostgreSQL database.
+
 ```yaml
 services:
   keycloak:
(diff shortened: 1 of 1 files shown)


thomasBousselin:
Hello @os-1693,
The KC_DB=postgres is added when we build the `easyglobalmarket/keycloak` image.
If you experienced an issue in your environment, it may be an issue on how we build the image.
The project building the image is not public yet. I will check if we can open it.

0x69d:
Hi @thomasBousselin,

Thank you for looking into this.

I just verified the behavior again in my local Docker environment using the `easyglobalmarket/keycloak:26.3.5` image with the exact `docker-compose.yml` provided in the docs.

When I run it **without** explicitly setting `KC_DB=postgres`, the Keycloak container starts, but the logs show that it falls back to the default H2 database:
```text
INFO  [io.quarkus] (main) Installed features: [agroal, cdi, hibernate-orm, jdbc-h2, keycloak, ...]
```
As a result, no data is actually persisted to the attached PostgreSQL container.

However, as soon as I explicitly add `- KC_DB=postgres` to the compose file, Keycloak correctly connects to PostgreSQL.

Even if the base image was intended to have this set during the build, it seems that running it (especially with `command: "start-dev"`) causes it to default back to H2 unless the environment variable is explicitly provided. Therefore, adding this to the documentation's example will prevent users from accidentally losing their data when they tear down the containers.

Let me know if you would like me to rename the branch (to pass the CI commit-check) and keep this PR!

thomasBousselin:
> Let me know if you would like me to rename the branch (to pass the CI commit-check) and keep this PR!

Yes let merge it. We will change it back if we fix the image.

thomasBousselin (approved):

PR #1648, opened 2026-03-05
fix: get entityTypeInfo returning two attributes with the same name if they have different types

thomasBousselin:


bobeal:
If I understand it correctly, it fixes the case where the same name of attribute is used by different types of attributes? If so, maybe change the PR title to make it clear :)

bobeal (approved):

PR #1684, opened 2026-03-27
feature: add error messages for partial attribute update errors

thomasBousselin:


bobeal:
@thomasBousselin your last pushed commits added a lot of unwanted commits

bobeal:
> @thomasBousselin your last pushed commits added a lot of unwanted commits

there is also a mismatch between the name of the branch (multi-values relationships) and the name of PR (error messages for partial attribute update)

thomasBousselin:
> @thomasBousselin your last pushed commits added a lot of unwanted commits

fixed

shared/src/main/kotlin/com/egm/stellio/shared/model/ExpandedMembers.kt
```diff
@@ -148,16 +152,20 @@ fun ExpandedAttributeInstance.getMemberValue(memberName: ExpandedTerm): Any? {
             it[JSONLD_VALUE_KW]
         }
     }
+    checkNotNull(value) {
```
bobeal: `checkNotNull` raises an `IllegalArgumentException` if `value` is null. Better use https://apidocs.arrow-kt.io/arrow-core/arrow.core.raise/ensure.html

shared/src/main/kotlin/com/egm/stellio/shared/model/ExpandedMembers.kt
```diff
@@ -148,16 +152,20 @@ fun ExpandedAttributeInstance.getMemberValue(memberName: ExpandedTerm): Any? {
...
+        BadRequestDataException(attributeMissingValueMessage(memberName, attributeName))
+    }
+    return value.right()
 }
 
-fun ExpandedAttributeInstance.getPropertyValue(): Any? =
-    getMemberValue(NGSILD_PROPERTY_VALUE)
+fun ExpandedAttributeInstance.getPropertyValue(attributeName: String = ""): Either<APIException, Any> =
```
bobeal: I did not find any calls passing an `attributeName`

shared/src/main/kotlin/com/egm/stellio/shared/model/NgsiLdEntity.kt
```diff
@@ -479,6 +479,7 @@ value class AttributeType(val uri: String)
 val NGSILD_PROPERTY_TYPE = AttributeType("https://uri.etsi.org/ngsi-ld/Property")
 val NGSILD_GEOPROPERTY_TYPE = AttributeType("https://uri.etsi.org/ngsi-ld/GeoProperty")
 val NGSILD_RELATIONSHIP_TYPE = AttributeType("https://uri.etsi.org/ngsi-ld/Relationship")
+val NGSILD_LIST_RELATIONSHIP_TYPE = AttributeType("https://uri.etsi.org/ngsi-ld/ListRelationship")
```
bobeal: stow passenger ;)

shared/src/main/kotlin/com/egm/stellio/shared/util/ErrorMessages.kt
```diff
@@ -125,6 +125,9 @@ object ErrorMessages {
         fun attributeCannotGetValueMessage(attribute: String) = "Cannot get value from attribute: $attribute"
 
         fun propertyMissingValueMessage(name: String) = "Property $name has an instance without a value"
+        fun attributeMissingValueMessage(memberValue: String, attributeName: String = "") =
```
bobeal: I don't think the message applies to the 2 places where it is used:
- The first place is related to the presence of an expected sub-attribute (for instance, a `value` sub-attribute if it is a property)
- The second place is related to a null value for a sub-attribute (btw, since it is expanded, I wonder how this can happen)

So 2 distinct errors.

bobeal (commented):
It is missing a few tests (here and / or in the Test Suite).

shared/src/main/kotlin/com/egm/stellio/shared/util/ErrorMessages.kt
thomasBousselin: I don't think the second message can happen unless someone remove something directly from the database.
Anyway I updated the message to be more specific.

shared/src/main/kotlin/com/egm/stellio/shared/model/ExpandedMembers.kt
```diff
@@ -109,55 +112,68 @@ fun ExpandedAttributeInstance.addSysAttrs(
...
-            finalValueType != null -> {
-                val finalValue = firstListEntry[JSONLD_VALUE_KW]
-                when (finalValueType) {
-                    NGSILD_DATE_TIME_TYPE -> ZonedDateTime.parse(finalValue as String)
-                    NGSILD_DATE_TYPE -> LocalDate.parse(finalValue as String)
-                    NGSILD_TIME_TYPE -> LocalTime.parse(finalValue as String)
-                    else -> finalValue
+fun ExpandedAttributeInstance.getMemberValue(
+    memberName: ExpandedTerm,
+    attributeName: String = ""
```
bobeal: Looking at usages of this function, `attributeName` is never used to provide the name of the "containing" attribute.

bobeal: `attributeName` is only used to display a nice error message. But that will fail the purpose if `attributeName` is empty.

shared/src/main/kotlin/com/egm/stellio/shared/model/ExpandedMembers.kt
```diff
@@ -109,55 +112,68 @@ fun ExpandedAttributeInstance.addSysAttrs(
...
+        val intermediateList = attributeInstance[memberName] as List<Map<String, Any>>
+        val value = if (intermediateList.size == 1) {
+            val firstListEntry = intermediateList[0]
+            val finalValueType = firstListEntry[JSONLD_TYPE_KW]
+            when {
+                finalValueType != null -> {
+                    val finalValue = firstListEntry[JSONLD_VALUE_KW]
+                    ensureNotNull(finalValue) {
+                        BadRequestDataException(
+                            attributeMissingValueMessage("$memberName[0].$JSONLD_VALUE_KW", attributeName)
```
bobeal: I think it misses some `{}` for the 1st parameter

shared/src/main/kotlin/com/egm/stellio/shared/util/ErrorMessages.kt
```diff
@@ -147,6 +150,8 @@ object ErrorMessages {
 
         const val CANNOT_ADD_SUBATTRIBUTE_MESSAGE = "Cannot add a sub-attribute into empty or multi-instance attribute"
         const val EXPECTED_SINGLE_ENTRY_MESSAGE = "Expected a single entry but got none or more than one"
+        const val NOT_IMPLEMENTED_PARTIAL_ATTRIBUTE_MESSAGE =
+            "Partial attribute update currently require the presence of the main subattribute"
```
bobeal:
```suggestion
            "Partial attribute update currently requires the presence of the main subattribute"
```

bobeal: "main subattribute" is not a notion defined in the specification. "... requires providing the value (in case of a Property) or object (in case of a Relationship)"?

shared/src/main/kotlin/com/egm/stellio/shared/model/ExpandedMembers.kt
thomasBousselin: the "[0]" is part of the error message as the value is missing from the first element of the list.

shared/src/main/kotlin/com/egm/stellio/shared/model/ExpandedMembers.kt
thomasBousselin: Since it is defaulted to "" it will still display a well-formed error message

shared/src/main/kotlin/com/egm/stellio/shared/util/ErrorMessages.kt
thomasBousselin: Yes, but it is better than having a partial list of what is missing. And i don't want to list every possible key.

shared/src/main/kotlin/com/egm/stellio/shared/model/ExpandedMembers.kt
bobeal: ah, yes.

shared/src/main/kotlin/com/egm/stellio/shared/util/ErrorMessages.kt
bobeal: Strictly speaking, every XXXProperty is a Property.

shared/src/main/kotlin/com/egm/stellio/shared/model/ExpandedMembers.kt
bobeal: Reformulating it: why adding a parameter (`attributeName`) which is never provided by callers (so always defaults to the empty string).

bobeal (approved):

Issue #1144, opened 2024-04-25
Refactor and finalize support for the NGSI-LD Query Language (4.9)


bobeal:
- Parsing of `q` queries is done via a regex, do a more bulletproof parsing based on the ABNF grammar defined in the specification (using antlr for instance)
- Add support for missing parts (`expandValues` and `jsonKeys`, seed target value, specific processing for `JsonProperty` and `LanguageProperty`,...)

PR #1697, opened 2026-04-08
fix: handle commas inside quoted values

bobeal:


shared/src/main/kotlin/com/egm/stellio/shared/util/QueryUtils.kt
```diff
@@ -52,12 +52,37 @@ fun String.isRange(): Boolean =
...
+            ',' if !inQuotes -> return true
+        }
+    }
+    return false
+}
 
-fun String.listOfValues(): Set<String> =
-    this.split(",").toSet()
+fun String.listOfValues(): Set<String> {
+    val values = mutableListOf<String>()
```
thomasBousselin, line 67: You could use a regex

bobeal: really?

bobeal: anyway, at the end, this is how the whole query language parsing will end up, so let's get used to this now.

shared/src/main/kotlin/com/egm/stellio/shared/util/QueryUtils.kt
```diff
@@ -52,12 +52,37 @@ fun String.isRange(): Boolean =
 fun String.rangeInterval(): Pair<Any, Any> =
     Pair(this.split("..")[0], this.split("..")[1])
 
-// a string value could contain a comma ... to be improved
-fun String.isValueList(): Boolean =
-    this.contains(",")
+fun String.isValueList(): Boolean {
```
thomasBousselin, line 55: You could also use a regex. (Not sure what will be most efficient)

shared/src/main/kotlin/com/egm/stellio/shared/util/QueryUtils.kt
thomasBousselin: Something like `regex.findAll(\"[^"]*"|[^,]+\)` should work. I think that using as much regex as possible in the query language will really help with code readability (and length).

bobeal: it introduces new problems for list of things that are not strings. since both solutions are not the perfect ones, let's close the discussion for now and follow up in #1144?

thomasBousselin (approved):

PR #1704, opened 2026-04-12
fix: ensure correct entity in notification when LER is asked

### LINKED ISSUE #1703: Incorrect entity payload in notifications with linked entities
When a notification is configured to send linked entities along with the entity that triggered the notification, the subscription service queries the search service to retrieve the entity with the linked entities. However, between the reception of the event and the call to the search service, the entity may have changed (typically if there are many updates sent at once) and thus the entity that is sent is not the correct one.

bobeal:


thomasBousselin (approved):
It does not fix the case where the linked entities have changed. But fixing this problem entirely will ask way bigger effort. So let's go with this for now.

PR #1712, opened 2026-04-22
feat: add distributed operation for Temporal Consumption endpoints

bobeal:


search-service/src/main/kotlin/com/egm/stellio/search/csr/model/Operation.kt
```diff
@@ -15,6 +15,9 @@ enum class Operation(
     @JsonProperty("federationOps")
     FEDERATION_OPS("federationOps"),
 
+    @JsonProperty("associationOps")
+    ASSOCIATION_OPS("associationOps"),
```
thomasBousselin, line 19: What is it?

search-service/src/main/kotlin/com/egm/stellio/search/csr/util/ContextSourceUtils.kt
```diff
@@ -138,7 +147,9 @@ object ContextSourceUtils {
         if (values.size == 1) values[0] else values
     }
 
-    // only meant to work with attributes under the normalized representation
+    // only meant to work with attributes under:
+    // - the normalized representation when retrieving or querying entities
+    // - the simplified or aggregated representation when retrieving or querying tmeporal entities
```
thomasBousselin: tmeporal

search-service/src/main/kotlin/com/egm/stellio/search/csr/service/DistributedTemporalEntityConsumptionService.kt
```diff
@@ -0,0 +1,173 @@
...
+        csr: SingleEntityInfoCSR,
+        csrFilters: CSRFilters,
+        path: String,
+        params: MultiValueMap<String, String>
+    ): Either<NGSILDWarning, Pair<String?, ClientResponse.Headers>> = either {
+        val contexts = getContextFromLinkHeaderOrDefault(httpHeaders, applicationProperties.contexts.core).getOrNull()!!
+        val uri = URI("${csr.endpoint}$path")
+
+        val queryParams = CollectionUtils.toMultiValueMap(params.toMutableMap())
+        queryParams.remove(QueryParameter.GEOMETRY_PROPERTY.key)
```
thomasBousselin: The removal of the GEOMETRY_PROPERTY was to avoid fetching data in geojson. It does not make sense for endpoints that does not support geo-json.

search-service/src/main/kotlin/com/egm/stellio/search/csr/util/ContextSourceUtils.kt
```diff
@@ -169,4 +180,144 @@ object ContextSourceUtils {
         this?.isDateTime() == true &&
             date?.isDateTime() == true &&
             ZonedDateTime.parse(this) < ZonedDateTime.parse(date)
+
+    fun mergeTemporalEntitiesLists(
+        localEntities: List<CompactedEntity>,
+        remoteEntitiesWithCSR: List<CompactedEntitiesWithCSR>,
+        temporalEntitiesQuery: TemporalEntitiesQueryFromGet
+    ): IorNel<NGSILDWarning, List<CompactedEntity>> {
+        val mergedEntityMap = localEntities.map { it.toMutableMap() }.associateBy { it[NGSILD_ID_TERM] }.toMutableMap()
```
thomasBousselin: EntityMap term correspond to something else in the specification.

search-service/src/main/kotlin/com/egm/stellio/search/csr/util/ContextSourceUtils.kt
```diff
@@ -169,4 +180,144 @@ object ContextSourceUtils {
...
+                getMergeTemporalNewValues(mergedEntity, entity, temporalEntitiesQuery, csr)
+                    .onRight { mergedEntity.putAll(it) }.leftOrNull()
+            }.toNonEmptyListOrNull()
+
+        return if (warnings == null)
+            Ior.Right(mergedEntity)
+        else Ior.Both(warnings, mergedEntity)
+    }
+
+    internal fun getMergeTemporalNewValues(
```
thomasBousselin, line 227: I don't find the name really clear here are some ideas : (no perfect one for me)
`CompactedEntity.mergeTemporalWithRemoteEntity`
`mergeTemporalEntity`
`mergeTwoTemporalEntities ` (with `mergeTemporalEntities `renamed `mergeAllEntities`).

search-service/src/main/kotlin/com/egm/stellio/search/csr/util/ContextSourceUtils.kt
```diff
@@ -169,4 +180,144 @@ object ContextSourceUtils {
...
+        temporalEntitiesQuery: TemporalEntitiesQuery,
+        csr: ContextSourceRegistration
+    ): Either<NGSILDWarning, Any> = either {
+        when (temporalEntitiesQuery.temporalRepresentation) {
+            TemporalRepresentation.NORMALIZED -> {
+                val currentInstances = toListOfNormalizedInstances(currentAttribute, csr).bind()
+                val remoteInstances = toListOfNormalizedInstances(remoteAttribute, csr).bind()
+                (currentInstances + remoteInstances).sortedBy {
+                    // order instances by the asked time property
+                    it[temporalEntitiesQuery.temporalQuery.timeproperty.propertyName] as String
```
thomasBousselin: If two attributes have the same time and datasetId? (same for temporalValues but basically "impossible" for aggregation)
I feel that it may happen quite a lot in the inclusive scenario where the data is store locally and remotely. All the point will be duplicated.

thomasBousselin (commented):
First review (without pagination)

search-service/src/main/kotlin/com/egm/stellio/search/csr/model/Operation.kt
bobeal: Looks like a display bug in the PDF of the spec:

image

Looking at table 4.20-2 in https://cim.etsi.org/NGSI-LD/official/clause-4.html#4.20, it is just a separator?

But looking at the Markdown for 2.0, it is back:

```
+-------------------------------------+-------------------------------------+
| Operation Group name                | Implements                          |
+=====================================+=====================================+
| federationOps                       | - retrieveEntity                    |
|                                     | - queryEntity                       |
|                                     | - queryBatch                        |
|                                     | - retrieveEntityTypes               |
|                                     | - retrieveEntityTypeDetails         |
|                                     | - retrieveEntityTypeInfo            |
|                                     | - retrieveAttrTypes                 |
|                                     | - retrieveAttrTypeDetails           |
|                                     | - retrieveAttrTypeInfo              |
|                                     | - createSubscription                |
|                                     | - updateSubscription                |
|                                     | - retrieveSubscription              |
|                                     | - querySubscription                 |
|                                     | - deleteSubscription                |
|                                     | - retrieveEntityMap                 |
|                                     | - updateEntityMap                   |
|                                     | - deleteEntityMap                   |
|                                     | - createEntityMapQueryEntity        |
|                                     | - retrieveContextSourceIdentity     |
+-------------------------------------+-------------------------------------+
| associationOps                      | - retrieveEntity                    |
|                                     | - queryEntity                       |
|                                     | - queryBatch                        |
|                                     | - retrieveEntityTypes               |
|                                     | - retrieveEntityTypeDetails         |
|                                     | - retrieveEntityTypeInfo            |
|                                     | - retrieveAttrTypes                 |
|                                     | - retrieveAttrTypeDetails           |
|                                     | - retrieveAttrTypeInfo              |
|                                     | - createSubscription                |
|                                     | - updateSubscription                |
|                                     | - retrieveSubscription              |
|                                     | - querySubscription                 |
|                                     | - deleteSubscription                |
|                                     | - retrieveContextSourceIdentity     |
```

search-service/src/main/kotlin/com/egm/stellio/search/csr/service/DistributedTemporalEntityConsumptionService.kt
bobeal: Indeed. Removed.

search-service/src/main/kotlin/com/egm/stellio/search/csr/util/ContextSourceUtils.kt
bobeal: Fixed.

search-service/src/main/kotlin/com/egm/stellio/search/csr/util/ContextSourceUtils.kt
bobeal: yeah... renamed to `mergedEntities` (echoes the `mergedEntity` from `mergeTemporalEntities()`.

search-service/src/main/kotlin/com/egm/stellio/search/csr/util/ContextSourceUtils.kt
bobeal: well, I basically derived it from the existing naming of `getMergeNewValues()`...

since it is similar to a reduce operation, it could be named something like `reduceRemoteTemporalEntity`

search-service/src/main/kotlin/com/egm/stellio/search/csr/util/ContextSourceUtils.kt
bobeal: I gonna add a specific case for inclusive CSRs.

But I don't get why it is almost impossible for aggregations?

thomasBousselin: Not impossible to happen, but impossible to implement, max and min could work but count and distinctCount you can't really know.

bobeal: ah, this, I agree.

search-service/src/main/kotlin/com/egm/stellio/search/csr/util/ContextSourceUtils.kt
thomasBousselin: Oh, I mixed up what the function was actually doing. I think original naming is fine.

search-service/src/main/kotlin/com/egm/stellio/search/csr/util/ContextSourceUtils.kt
bobeal: Looking again at the spec, it also misses a specific handling for auxiliary context sources:

```
For any auxiliary Context Source Registrations the remote Attribute data received is added
to S1 only when the Attribute instance, whose value of the timeproperty, which is used for the temporal
query (observedAt as default), is not present in any of the Attribute instances received from elsewhere
```

bobeal: The full rules:

```
- For any exclusive, redirect and inclusive Context Source, the request is forwarded for remote
retrieval by matching endpoints. and remote Attribute data for the Entity is received. The result is then
merged together with S1 according to the algorithm defined in clause 4.5.5.
- For any auxiliary Context Source Registrations the remote Attribute data received is added
to S1 only when the Attribute instance, whose value of the timeproperty, which is used for the temporal
query (observedAt as default), is not present in any of the Attribute instances received from elsewhere.
```

So the spec does not make any difference between exclusive, redirect and inclusive. BUT 4.5.5 does not apply to temporal attributes instances since it refers to "the most current Attribute instance". It could only be applied if we ask for the normalized representation with `sysAttrs` to the Context Sources (whatever is asked for in the query), then apply the 4.5.5 rule on each temporal instance using `modifiedAt`, and then do the `temporalValues` / `aggregatedValues` representation stuff locally. That's a big change.

thomasBousselin: To me it can all be simplified by doing :  only keep the existing value when you have the same timestamp (and datasetId).

That way auxiliary data will never be added if there is already an instance for the timestamp.

I also have an example if you need convincing for the redirect case: 
- If you have two redirect CSR on the same attribute, you will send every provision to both CSR and will always have duplicated data. 

For exclusive, it can't happen so no need to take them into account.

bobeal: > To me it can all be simplified in : `only keep the existing value when you have the same timestamp (and datasetId).`

you meant "always keep the existing value when you have the same timestamp (and datasetId)." ?

thomasBousselin: Unless you want to ask for sysAttrs and do the merge of value with it but i am not sure it is realisitic from a performance point of vue.

thomasBousselin: >you meant "always keep the existing value when you have the same timestamp (and datasetId)." ?

yes

bobeal: That's an acceptable compromise for the common use-cases. It will not handle super exotic (weird?) use cases, but we can survive with this.

search-service/src/test/kotlin/com/egm/stellio/search/csr/util/ContextSourceUtilsTemporalTests.kt
```diff
@@ -0,0 +1,303 @@
...
+import org.junit.jupiter.api.Assertions.assertNotNull
+import org.junit.jupiter.api.Assertions.assertNull
+import org.junit.jupiter.api.Assertions.assertTrue
+import org.junit.jupiter.api.Test
+import org.springframework.test.context.ActiveProfiles
+
+@ActiveProfiles("test")
+class ContextSourceUtilsTemporalTests {
+
+    private val localIncomingNormalized =
```
thomasBousselin, line 28: Creating that many new files will make it hard to work on it later.

search-service/src/main/kotlin/com/egm/stellio/search/csr/util/ContextSourceUtils.kt
bobeal: Implemented

search-service/src/test/kotlin/com/egm/stellio/search/csr/util/ContextSourceUtilsTemporalTests.kt
bobeal: Yes and no, both ways have drawbacks.

Using such files, you are closer (equal) to what the app will handle IRL. It is also easier to visualize inputs and expectations. And it reduces a lot the boilerplate code in the tests. Yes, that's a lot of files but I don't expect a lot of maintenance on them (and don't know what you really mean by "work on it later")?

And creating lists of lists of lists in tests cases is not so fun and you can quickly end up with false inputs without ever noticing it.

thomasBousselin: If you want to keep it. They should all be in a separate subDirectory to simplify browsing the files.

thomasBousselin (commented):
I have started checking on the capabilities on TreeSet and I think it would help performances if we treated all the temporal instances as a TreeSet (could also help for other performance gain). Anyway event though I want the distributed temporal operation to have the best performance as possible. I think we can keep this subject for another time.

thomasBousselin (approved):

PR #1726, opened 2026-05-04
fix: disable by default export of metrics

bobeal:


thomasBousselin (commented):
before you needed to add one configuration now you need two?
Can we make the loading of the  bin depend on the new variable?

bobeal:
> before you needed to add one configuration now you need two?

Unfortunately, yes, it seems so. But before, it was only about logs. Now, it is about logs and metrics.

> Can we make the loading of the bin depend on the new variable?

The loading of the bin? What do you mean?

thomasBousselin (approved):
I was thinking of the `InstallOpenTelemetryAppender` but it is not as simple as i was thinking because of the `logback-spring.xml`.

PR #1728, opened 2026-05-04
feat: add support for contextSourceInfo field in CSR

### LINKED ISSUE #1713: Add support for the `contextSourceInfo` field in CSR
See description in https://cim.etsi.org/NGSI-LD/official/clause-5.html#5.2.9

bobeal:


search-service/src/main/kotlin/com/egm/stellio/search/csr/model/ContextSourceInfo.kt
```diff
@@ -0,0 +1,3 @@
+package com.egm.stellio.search.csr.model
+
+data class ContextSourceInfo(val key: String, val value: String)
```
thomasBousselin: Is it better than using a Pair?

search-service/src/main/kotlin/com/egm/stellio/search/csr/model/ContextSourceRegistration.kt
```diff
@@ -152,6 +154,12 @@ data class ContextSourceRegistration(
                 { it.right() },
                 { it.toAPIException(csrFailedToParseMessage(it.message)).left() }
             )
+
+        fun deserializeContextSourceInfo(input: String?): List<ContextSourceInfo>? {
```
thomasBousselin: I would also use a Map<String, String> instead of a List of ContextSourceInfo

search-service/src/main/kotlin/com/egm/stellio/search/csr/model/ContextSourceInfo.kt
bobeal: aligned with https://github.com/stellio-hub/stellio-context-broker/blob/develop/subscription-service/src/main/kotlin/com/egm/stellio/subscription/model/EndpointInfo.kt.

another PR could focus on improving this.

search-service/src/main/kotlin/com/egm/stellio/search/csr/model/ContextSourceRegistration.kt
bobeal: same here, it mimics what is done for `EndpointInfo`

thomasBousselin (dismissed):

PR #1739, opened 2026-05-13
feat: add more detailed Claude configuration

bobeal:
TBH, I think there is a bit too much files introduced in this PR and that some of them can be removed. But let's talk about it.


bobeal:
@AdamLewis005 also have a look at existing files in the *.claude* directory.

PR #1745, opened 2026-05-19
fix(csr): persist and retrieve CSR status

bobeal:


search-service/src/main/kotlin/com/egm/stellio/search/csr/service/ContextSourceRegistrationService.kt
```diff
@@ -222,7 +233,11 @@ class ContextSourceRegistrationService(
                 context_source_info,
                 tenant,
                 created_at,
-                modified_at
+                modified_at,
+                status,
+                times_sent,
+                times_failed,
+                last_failure
```
thomasBousselin, line 240: Are users supposed to modify the status?

bobeal: no, but these lines are in the `getContextSourceRegistrations` function so I don't get the point

thomasBousselin: This comment was meant for the upsert changes

bobeal: ah. no, users can't modify the `status`.

PR #1746, opened 2026-05-21
feat: apply temporal pagination only locally

thomasBousselin:
This is a hard revert from the commit moving the Content range from the service to the handler for the distribution Merge request.
The new pagination was not entirely working with the distribution.

And calculating pagination is very costly on CompactedEntity since you can't use the size of the attribute list
 (because datasetIds and temporal instance are merged in a single list) to know if the pagination is triggered or not.


search-service/src/main/kotlin/com/egm/stellio/search/temporal/util/TemporalPaginationUtils.kt
```diff
@@ -1,185 +1,89 @@
...
-import com.egm.stellio.shared.model.COMPACTED_ENTITY_CORE_MEMBERS
-import com.egm.stellio.shared.model.CompactedAttributeInstance
-import com.egm.stellio.shared.model.CompactedEntity
-import com.egm.stellio.shared.model.TEMPORAL_REPRESENTATION_TERMS
 import java.time.ZonedDateTime
 
 typealias Range = Pair<ZonedDateTime, ZonedDateTime>
 
-object TemporalPaginationUtils {
+object TemporalPaginationService {
```
bobeal: according to project conventions, *Service is reserved for Spring @Service-managed beans.

search-service/src/test/kotlin/com/egm/stellio/search/temporal/util/TemporalPaginationUtilsTests.kt
```diff
@@ -1,117 +1,145 @@
...
 import org.junit.jupiter.api.Test
+import org.springframework.boot.context.properties.EnableConfigurationProperties
+import org.springframework.boot.test.context.SpringBootTest
 import org.springframework.test.context.ActiveProfiles
 import java.time.ZonedDateTime
+import java.util.UUID
 
 @ActiveProfiles("test")
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.NONE, classes = [TemporalPaginationService::class])
+@EnableConfigurationProperties(SearchProperties::class)
```
bobeal: you don't need these 2 lines for this test. `TemporalPagination[Utils|Service]` is a plain Kotlin object.

search-service/src/test/kotlin/com/egm/stellio/search/temporal/util/TemporalPaginationUtilsTests.kt
```diff
@@ -1,117 +1,145 @@
...
+        withAudit = false
+    )
 
     @Test
-    fun `calculateRangeFromCompactedEntity should return timeAt as range-start for BETWEEN timerel`() {
-        val entity = mapOf(
-            NGSILD_ID_TERM to entityId,
-            INCOMING_TERM to (0..4).map { normalizedInstance(leastRecentTimestamp.plusMinutes(it.toLong())) }
-        )
+    fun `range calculation with timerel between should return range-start = timeAt`() = runTest {
```
bobeal: all the `runTest {}` are unnecessary in this test class since no `suspend` functions are called.

search-service/src/main/kotlin/com/egm/stellio/search/temporal/util/TemporalPaginationUtils.kt
```diff
@@ -1,185 +1,89 @@
...
+        attributesWithInstances: AttributesWithInstances,
+        range: Range,
+    ): AttributesWithInstances =
+        attributesWithInstances.mapValues { (_, value) ->
+            value.filter { range.contain(it.getComparableTime()) }
         }
 
-    private fun Range.contains(time: ZonedDateTime): Boolean =
-        time in second..first || time in first..second
+    private fun Range.contain(time: ZonedDateTime): Boolean =
```
bobeal:
```suggestion
    private fun Range.contains(time: ZonedDateTime): Boolean =
```

bobeal (commented):
- Update the title of the PR (correct form and more descriptive)
- It is a good opportunity to add a limitations page in the docs to explain limitations on currently implemented features. In the current case, it would explain that temporal pagination is only applied to local entities and that temporal data from context sources is blindly added to the result

docs/user/API_walkthrough.md
```diff
@@ -955,6 +955,20 @@ Note 2: You must provide a time filter (`timerel`, `timeAt`) and an entity filte
 The list contains multiple values if multiple entities match the request.
 </details>
 
+### Pagination of temporal data
```
bobeal: Would not it be better to have a dedicated page to list these limitations?

I could for instance add in it what is not implemented for the NGSI-LD Query Language. And I am pretty sure there are others to add.

More over, the API walkthrough is a bit early to start talking about distributed temporal queries :)

bobeal:
Don't forget the easy to fix Sonar issue.

docs/user/API_walkthrough.md
thomasBousselin: It depends on how you see the API_walkthrough. Today it is at the same time the documentation from each endpoint and a tutorial on how to use the API.
I think having the two together can be a good idea. (easier to maintain and to find what you want)
I can add this in another page and add a link to it here. Or i can put it in a <details>Yes a detail like that thanks github</details> to avoid confusing the users too much.

bobeal: Hmm, OK, let's keep it in the walkthrough for now.

bobeal (dismissed):

thomasBousselin:
The only new changes are to put the pagination doc inside a beautiful details.

bobeal (approved):

PR #1753, opened 2026-05-26
feat: implement a Kafka listener to receive telemetry data

### LINKED ISSUE #1731: Ingestion of data through Kafka (events)
As I did understud base on documentation we need to do this steps:

1. Sensor sends new value
2. IngestionService receives it
3. IngestionService calls Stellio:
   GET /ngsi-ld/v1/entities/{entityId}
4. service patches the attribute locally in that JSON
5. service puts that result into updatedEntity, and updates other properties (like previousPayload etc)
6. service publishes the Kafka ATTRIBUTE_UPDATE event

IngestionService => is one of our services that receive the events and sends to Kafka (in our case EventHub)

question:
What will be the benefit if we use this instead of Http requests to Stellio having in mind that we need to make another request to Stellio plus do the patching of properties manual ?

note:
"previousPayload" is not mentioned on the documentation (https://stellio.readthedocs.io/en/latest/admin/architecture.html) but if we make a request without it SubscriptionService throws exception

`2026-05-05 06:38:42,781 [atcher-worker-4] WARN  c.e.s.s.l.EntityEventListenerService     - dispatchEntityEvent$suspendImpl - Entity event processing has failed: Instantiation of [simple type, class com.egm.stellio.shared.model.AttributeUpdateEvent] value failed for JSON property previousPayload due to missing (therefore NULL) value for creator parameter previousPayload which is a non-nullable type

 at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); byte offset: #UNKNOWN] (through reference chain: com.egm.stellio.shared.model.AttributeUpdateEvent["previousPayload"])

 tools.jackson.module.kotlin.KotlinInvalidNullException: Instantiation of [simple type, class com.egm.stellio.shared.model.AttributeUpdateEvent] value failed for JSON property previousPayload due to missing (therefore NULL) value for creator parameter previousPayload which is a non-nullable type

 at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); byte offset: #UNKNOWN] (through reference chain: com.egm.stellio.shared.model.AttributeUpdateEvent["previousPayload"])`

bobeal:
- Introduces TelemetryDataMessage as the message contract
- Adds TelemetryListener that consumes messages from the `cim.telemetry` Kafka topic
- Adds EntityService.mergeAttribute — a lightweight entry point that builds an expanded attribute fragment, delegates to the existing mergeAttributes pipeline, and publishes an event on `cim.entity._CatchAll`
- Updates the architecture documentation with the topic contract and field reference

### DIFF
--- a/docs/admin/architecture.md
+++ b/docs/admin/architecture.md
@@ -113,38 +113,6 @@ are JSON-LD compacted, but when sent by Stellio, they are JSON-LD expanded):
 }
 ```
 
-### Telemetry ingestion via Kafka
-
-For high-throughput IoT scenarios, attribute values can be pushed directly to Kafka without going through the HTTP API.
-This avoids the overhead of HTTP connection handling and offers an improved resilience.
-
-**Topic**: `cim.telemetry`
-
-**Constraints**:
-- The entity must already exist (provisioned once via HTTP).
-- Only `Property` values are supported; other attribute types must use the HTTP API.
-- No authentication — Kafka is a trusted, internal channel.
-
-**Message format** (the message key must be set to the `entityId`):
-
-```json
-{
-  "tenantName": "urn:ngsi-ld:tenant:stellio",
-  "entityId": "urn:ngsi-ld:Vehicle:A4567",
-  "attributeName": "https://vocab.egm.io/speed",
-  "datasetId": "urn:ngsi-ld:Dataset:GPS",
-  "value": 25.3,
-  "observedAt": "2026-01-01T00:00:00Z"
-}
-```
-
-On receipt, the search-service:
-- creates the attribute if it does not exist yet.
-- applies merge semantics (NGSI-LD §5.5.12), preserving existing attribute metadata, if it exists.
-
-An `ATTRIBUTE_CREATE` or `ATTRIBUTE_UPDATE` event is then published to `cim.entity._CatchAll` so subscriptions fire as
-normal.
-
 ### Mapping from Core API operations to `notificationTrigger` in subscriptions
 
 Starting from version 1.6.1 of the NGSI-LD specification, subscriptions support a new `notificationTrigger` member (see 5.2.12 for more details).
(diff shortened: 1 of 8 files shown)


docs/admin/architecture.md
```diff
@@ -113,6 +113,38 @@ are JSON-LD compacted, but when sent by Stellio, they are JSON-LD expanded):
 }
 ```
 
+### Telemetry ingestion via Kafka
```
thomasBousselin: If you look are looking high performance data ingestion not sure you will check inside architecture to find your result. It may be better to put this in a separate document. (I am also not sure if it is pertinent to keep the description of the purely internal notification event.)

search-service/src/main/kotlin/com/egm/stellio/search/entity/model/TelemetryDataMessage.kt
```diff
@@ -0,0 +1,15 @@
...
+import java.net.URI
+import java.time.ZonedDateTime
+
+data class TelemetryDataMessage(
+    val tenantName: String = DEFAULT_TENANT_NAME,
+    val entityId: URI,
+    val attributeName: ExpandedTerm,
+    val datasetId: URI? = null,
+    val value: Any,
+    val observedAt: ZonedDateTime
```
thomasBousselin, line 14: Can we make observedAt optionnal?

bobeal: Hmm, no, that would break the whole point of this feature, which is to receive telemetry data.

docs/admin/architecture.md
bobeal: indeed, moved to a new page (in user section since it is more for programmers than for administrators).

thomasBousselin (dismissed):

PR #1764, opened 2026-06-05
feat(core): implement missing behaviours in handling of NGSI-LD Nulls

### LINKED ISSUE #1589: Implement missing behaviors in handling of NGSI-LD Nulls
- it should return a `BadRequestData` error if used in other contexts that 5.5.8 and 5.5.12
  - Add a check in `NgsiLdXXXProperty.create()` in `NgsiLdEntity`?
- it should allow to delete a sub-Attribute

bobeal:
Per ETSI GS CIM 009 §4.5.18, §5.5.8, and §5.5.12:

- reject `urn:ngsi-ld:null` in create, append, and replace operations (`BadRequestData`)
- reject `urn:ngsi-ld:null` in `datasetId` member for merge patch and partial attribute update
- remove sub-attributes whose value is `urn:ngsi-ld:null` in merge patch (5.5.12)
- remove sub-attributes whose value is `urn:ngsi-ld:null` in partial attribute update (5.5.8)
- merge `LanguageProperty` languageMap entries and remove `@none` language via `urn:ngsi-ld:null` (§4.5.18)
- remove keys from `Property` object values and `JsonProperty` json objects via `urn:ngsi-ld:null` (§5.5.12 example 2)

Limitations:
- Removing an `observedAt` in a merge+patch operation is not possible since the value is transformed in a ZonedDateTime before applying the merge (and Stellio then returns a 400 if `observedAt` is `urn:ngsi-ld:null`). Allowing this use-case implies many changes and is thus postponed until it is strongly needed.

### DIFF
--- a/search-service/src/main/kotlin/com/egm/stellio/search/entity/util/AttributeUtils.kt
+++ b/search-service/src/main/kotlin/com/egm/stellio/search/entity/util/AttributeUtils.kt
@@ -272,7 +272,7 @@ fun mergePatch(
             isNgsiLdNullDatasetId(attrName, attrValue) ->
                 raise(BadRequestDataException(NGSI_LD_NULL_NOT_ALLOWED_IN_DATASET_ID_MESSAGE))
             attrName == NGSILD_LANGUAGEPROPERTY_LANGUAGEMAP ->
-                target[attrName] = mergeLanguageMap(source, attrValue as List<Map<String, String>>)
+                target[attrName] = mergeLanguageMap(source, attrValue as ExpandedLanguageMapValue)
             isNgsiLdNullSubAttribute(attrValue) || isNgsiLdNullValue(attrValue) ->
                 target.remove(attrName)
             !source.containsKey(attrName) ->
--- a/search-service/src/main/kotlin/com/egm/stellio/search/entity/util/AttributeUtils.kt
+++ b/search-service/src/main/kotlin/com/egm/stellio/search/entity/util/AttributeUtils.kt
@@ -13,6 +13,7 @@ import com.egm.stellio.search.entity.model.AttributeMetadata
 import com.egm.stellio.shared.model.APIException
 import com.egm.stellio.shared.model.BadRequestDataException
 import com.egm.stellio.shared.model.ExpandedAttributeInstance
+import com.egm.stellio.shared.model.ExpandedTerm
 import com.egm.stellio.shared.model.JSONLD_ID_KW
 import com.egm.stellio.shared.model.JSONLD_LANGUAGE_KW
 import com.egm.stellio.shared.model.JSONLD_TYPE_KW
@@ -21,7 +22,6 @@ import com.egm.stellio.shared.model.NGSILD_DATASET_ID_IRI
 import com.egm.stellio.shared.model.NGSILD_JSONPROPERTY_JSON
 import com.egm.stellio.shared.model.NGSILD_LANGUAGEPROPERTY_LANGUAGEMAP
 import com.egm.stellio.shared.model.NGSILD_NULL
-import com.egm.stellio.shared.model.NGSILD_PREFIX
 import com.egm.stellio.shared.model.NGSILD_PROPERTY_VALUE
 import com.egm.stellio.shared.model.NGSILD_VOCABPROPERTY_VOCAB
 import com.egm.stellio.shared.model.NgsiLdAttributeInstance
@@ -166,27 +166,27 @@ fun guessRelationshipValueType(
             Pair(Attribute.AttributeValueType.ARRAY, Triple(objectId.ids.asJsonB(), null, null))
     }
 
-private fun isNgsiLdNullSubAttribute(attrValue: Any): Boolean {
-    val instance = (attrValue as? List<*>)?.firstOrNull() as? ExpandedAttributeInstance ?: return false
+private fun isNgsiLdNullSubAttribute(attrValue: List<Any>): Boolean {
+    val instance = attrValue.firstOrNull() as? ExpandedAttributeInstance ?: return false
     val typeUri = (instance[JSONLD_TYPE_KW] as? List<*>)?.firstOrNull() as? String ?: return false
-    val attributeType = AttributeType.entries.find { typeUri == "$NGSILD_PREFIX${it.name}" } ?: return false
+    val attributeType = AttributeType.entries.find { typeUri == it.toExpandedName() } ?: return false
     return hasNgsiLdNullValue(instance, attributeType)
 }
 
-private fun isNgsiLdNullId(attrValue: Any): Boolean =
-    ((attrValue as? List<*>)?.firstOrNull() as? Map<*, *>)?.get(JSONLD_ID_KW) == NGSILD_NULL
+private fun isNgsiLdNullDatasetId(attrName: ExpandedTerm, attrValue: List<Any>): Boolean =
+    attrName == NGSILD_DATASET_ID_IRI &&
+        (attrValue.firstOrNull() as? Map<*, *>)?.get(JSONLD_ID_KW) == NGSILD_NULL
 
-private fun isNgsiLdNullValue(attrValue: Any): Boolean =
-    ((attrValue as? List<*>)?.firstOrNull() as? Map<*, *>)?.get(JSONLD_VALUE_KW) == NGSILD_NULL
+private fun isNgsiLdNullValue(attrValue: List<Any>): Boolean =
+    (attrValue.firstOrNull() as? Map<*, *>)?.get(JSONLD_VALUE_KW) == NGSILD_NULL
 
 private fun mergeLanguageMap(
     source: ExpandedAttributeInstance,
-    attrName: String,
-    attrValue: List<Any>
+    updatedLanguageMap: List<Map<String, String>>
 ): List<Any> {
-    val sourceLangEntries = source[attrName] as List<Map<String, String>>
+    val sourceLangEntries = source[NGSILD_LANGUAGEPROPERTY_LANGUAGEMAP] as List<Map<String, String>>
     val targetLangEntries = sourceLangEntries.toMutableList()
-    (attrValue as List<Map<String, String>>).forEach { langEntry ->
+    updatedLanguageMap.forEach { langEntry ->
         targetLangEntries.removeIf { it[JSONLD_LANGUAGE_KW] == langEntry[JSONLD_LANGUAGE_KW] }
         if (langEntry[JSONLD_VALUE_KW] != NGSILD_NULL)
             targetLangEntries.add(langEntry)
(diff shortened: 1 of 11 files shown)


search-service/src/main/kotlin/com/egm/stellio/search/entity/util/AttributeUtils.kt
```diff
@@ -143,6 +149,59 @@ fun guessPropertyValueType(
...
+    val instance = (attrValue as? List<*>)?.firstOrNull() as? ExpandedAttributeInstance ?: return false
+    val typeUri = (instance[JSONLD_TYPE_KW] as? List<*>)?.firstOrNull() as? String ?: return false
+    val attributeType = AttributeType.entries.find { typeUri == "$NGSILD_PREFIX${it.name}" } ?: return false
+    return hasNgsiLdNullValue(instance, attributeType)
+}
+
+private fun isNgsiLdNullId(attrValue: Any): Boolean =
+    ((attrValue as? List<*>)?.firstOrNull() as? Map<*, *>)?.get(JSONLD_ID_KW) == NGSILD_NULL
+
+private fun isNgsiLdNullValue(attrValue: Any): Boolean =
```
thomasBousselin: This function is only used in `isNgsiLdNullSubAttribute(attrValue) || isNgsiLdNullValue(attrValue)`. Is it usefull to have two separated functions.

search-service/src/main/kotlin/com/egm/stellio/search/entity/util/AttributeUtils.kt
```diff
@@ -143,6 +149,59 @@ fun guessPropertyValueType(
...
+    ((attrValue as? List<*>)?.firstOrNull() as? Map<*, *>)?.get(JSONLD_VALUE_KW) == NGSILD_NULL
```
thomasBousselin: The fact that these functions all take an Any in entry make them hard to understand.

search-service/src/main/kotlin/com/egm/stellio/search/entity/util/AttributeUtils.kt
```diff
@@ -143,6 +149,59 @@ fun guessPropertyValueType(
...
     }
 
+private fun isNgsiLdNullSubAttribute(attrValue: Any): Boolean {
+    val instance = (attrValue as? List<*>)?.firstOrNull() as? ExpandedAttributeInstance ?: return false
+    val typeUri = (instance[JSONLD_TYPE_KW] as? List<*>)?.firstOrNull() as? String ?: return false
+    val attributeType = AttributeType.entries.find { typeUri == "$NGSILD_PREFIX${it.name}" } ?: return false
+    return hasNgsiLdNullValue(instance, attributeType)
+}
+
+private fun isNgsiLdNullId(attrValue: Any): Boolean =
```
thomasBousselin: This function is only used to check datasetId is not equal to "urn:ngsi-ld-null".

search-service/src/main/kotlin/com/egm/stellio/search/entity/util/AttributeUtils.kt
```diff
@@ -143,6 +149,59 @@ fun guessPropertyValueType(
         else -> Pair(Attribute.AttributeValueType.STRING, Triple(value.toString(), null, null))
     }
 
+private fun isNgsiLdNullSubAttribute(attrValue: Any): Boolean {
+    val instance = (attrValue as? List<*>)?.firstOrNull() as? ExpandedAttributeInstance ?: return false
+    val typeUri = (instance[JSONLD_TYPE_KW] as? List<*>)?.firstOrNull() as? String ?: return false
+    val attributeType = AttributeType.entries.find { typeUri == "$NGSILD_PREFIX${it.name}" } ?: return false
```
thomasBousselin: use toExpandedName instead of this concatenation...

thomasBousselin: Shall it return an error if one of the cast fails?

search-service/src/main/kotlin/com/egm/stellio/search/entity/util/AttributeUtils.kt
```diff
@@ -165,54 +224,59 @@ fun Json.toExpandedAttributeInstance(): ExpandedAttributeInstance =
...
+): Either<APIException, Pair<String, ExpandedAttributeInstance>> = either {
     val target = source.toMutableMap()
     update.forEach { (attrName, attrValue) ->
-        if (!source.containsKey(attrName)) {
-            target[attrName] = attrValue
-        } else if (
+        when {
+            attrName == NGSILD_DATASET_ID_IRI && isNgsiLdNullId(attrValue) ->
+                raise(BadRequestDataException(NGSI_LD_NULL_NOT_ALLOWED_IN_DATASET_ID_MESSAGE))
+            listOf(NGSILD_LANGUAGEPROPERTY_LANGUAGEMAP).contains(attrName) ->
```
thomasBousselin: use attrName == NGSILD_LANGUAGEPROPERTY_LANGUAGEMAP  instead...

search-service/src/main/kotlin/com/egm/stellio/search/entity/util/AttributeUtils.kt
```diff
@@ -143,6 +149,59 @@ fun guessPropertyValueType(
...
+
+private fun isNgsiLdNullId(attrValue: Any): Boolean =
+    ((attrValue as? List<*>)?.firstOrNull() as? Map<*, *>)?.get(JSONLD_ID_KW) == NGSILD_NULL
+
+private fun isNgsiLdNullValue(attrValue: Any): Boolean =
+    ((attrValue as? List<*>)?.firstOrNull() as? Map<*, *>)?.get(JSONLD_VALUE_KW) == NGSILD_NULL
+
+private fun mergeLanguageMap(
+    source: ExpandedAttributeInstance,
+    attrName: String,
```
thomasBousselin: attrName will always be NGSILD_LANGUAGEPROPERTY_LANGUAGEMAP

search-service/src/main/kotlin/com/egm/stellio/search/entity/util/AttributeUtils.kt
```diff
@@ -143,6 +149,59 @@ fun guessPropertyValueType(
...
+    attrValue: List<Any>
```
thomasBousselin: Name (and maybe also the Type) could be more clear maybe `newLanguageMap`?

search-service/src/main/kotlin/com/egm/stellio/search/entity/util/AttributeUtils.kt
```diff
@@ -143,6 +149,59 @@ fun guessPropertyValueType(
...
+): List<Any> {
+    val sourceLangEntries = source[attrName] as List<Map<String, String>>
```
thomasBousselin: We could introduce a asLanguageMap() function and maybe a LanguageMap typealias.

search-service/src/main/kotlin/com/egm/stellio/search/entity/util/AttributeUtils.kt
```diff
@@ -143,6 +149,59 @@ fun guessPropertyValueType(
...
+/**
+ * Removes from the merged map any keys that have an NGSI-LD Null value in the update.
+ *
+ * Two representations are handled:
+ * - JsonProperty: value is `{"@value": {...raw JSON...}, "@type": "@json"}` — null keys are inside the @value map
+ * - Property with object value: expanded as nested JSON-LD with IRI keys — null keys are top-level entries whose
+ *   single expanded value is `[{"@value": "urn:ngsi-ld:null"}]`
+ */
+private fun applyNgsiLdNullRemoval(merged: Map<String, Any>, update: Any): Map<String, Any> {
+    val updateMap = update as? Map<*, *> ?: return merged
```
thomasBousselin: This will fail silently if someone call this function with the wrong arguments?

search-service/src/main/kotlin/com/egm/stellio/search/entity/util/AttributeUtils.kt
```diff
@@ -143,6 +149,59 @@ fun guessPropertyValueType(
...
+
+/**
+ * Removes from the merged map any keys that have an NGSI-LD Null value in the update.
+ *
+ * Two representations are handled:
+ * - JsonProperty: value is `{"@value": {...raw JSON...}, "@type": "@json"}` — null keys are inside the @value map
+ * - Property with object value: expanded as nested JSON-LD with IRI keys — null keys are top-level entries whose
+ *   single expanded value is `[{"@value": "urn:ngsi-ld:null"}]`
+ */
+private fun applyNgsiLdNullRemoval(merged: Map<String, Any>, update: Any): Map<String, Any> {
```
thomasBousselin: Variable names are unclear.

search-service/src/main/kotlin/com/egm/stellio/search/entity/util/AttributeUtils.kt
```diff
@@ -143,6 +149,59 @@ fun guessPropertyValueType(
...
+        val nullKeys = jsonContent.filter { (_, v) -> v == NGSILD_NULL }.keys
+        val mergedContent = merged[JSONLD_VALUE_KW] as? Map<String, Any>
+        return if (nullKeys.isEmpty() || mergedContent == null) merged
+        else merged + (JSONLD_VALUE_KW to mergedContent.filterKeys { it !in nullKeys })
+    }
+
+    val nullKeys = updateMap.filter { (_, v) ->
+        (v as? List<*>)?.singleOrNull()?.let { it as? Map<*, *> }?.get(JSONLD_VALUE_KW) == NGSILD_NULL
+    }.keys
+    return if (nullKeys.isEmpty()) merged else merged.filterKeys { it !in nullKeys }
```
thomasBousselin: First part of the function is only for jsonProperty and second par is only for Property.
It could be generic and receive an attrName. Or it should be two different functions.

search-service/src/main/kotlin/com/egm/stellio/search/entity/util/AttributeUtils.kt
bobeal: Yes, I prefer having two distinct functions. Small and focused. Also clear on the call site.

search-service/src/main/kotlin/com/egm/stellio/search/entity/util/AttributeUtils.kt
bobeal: Yes, hence the doc: "Removes from the **merged** map any keys that have an NGSI-LD Null value in the **update**"

No better idea for now.

search-service/src/main/kotlin/com/egm/stellio/search/entity/util/AttributeUtils.kt
bobeal: The first two can fail if not called for a sub-Attribute and that's not an error.

The third one can only fail for a partial attribute update (for a merge patch, the parsing will fail before reaching this function if the attribute type is not valid). But not really the job of this function to do validation, I prefer leaving this as it is.

search-service/src/main/kotlin/com/egm/stellio/search/entity/util/AttributeUtils.kt
bobeal: Don't see how it can be generic. But added `attrName` as a parameter to make it clearer.

search-service/src/main/kotlin/com/egm/stellio/search/entity/util/AttributeUtils.kt
```diff
@@ -165,54 +231,59 @@ fun Json.toExpandedAttributeInstance(): ExpandedAttributeInstance =
...
     val target = source.toMutableMap()
     update.forEach { (attrName, attrValue) ->
-        if (!source.containsKey(attrName)) {
-            target[attrName] = attrValue
-        } else if (
+        when {
+            isNgsiLdNullDatasetId(attrName, attrValue) ->
+                raise(BadRequestDataException(NGSI_LD_NULL_NOT_ALLOWED_IN_DATASET_ID_MESSAGE))
+            attrName == NGSILD_LANGUAGEPROPERTY_LANGUAGEMAP ->
+                target[attrName] = mergeLanguageMap(source, attrValue as List<Map<String, String>>)
```
thomasBousselin:
```suggestion
                target[attrName] = mergeLanguageMap(source, attrValue as ExpandedLanguageMapValue)
```

shared/src/main/kotlin/com/egm/stellio/shared/model/NgsiLdEntity.kt
thomasBousselin:
```suggestion
    val languageMap: ExpandedLanguageMapValue,
```

PR #1767, opened 2026-06-10
fix(core): support for partial attribute update without value

### LINKED ISSUE #1689: Partial attribute update without value
Sending a partial attribute update without his core sub attribute (value, object, json) return a BadRequestData.

For example : 
PATCH : http://localhost:8080/ngsi-ld/v1/entities/urn:1/attrs/property
```
{
    "unitcode": "m"
}
```
returns an error

when:

PATCH :
```
{
    "value": "a",
    "unitcode": "m"
}
```
returns a 204

The problem is the same for subattributes:
```
{
    "test": {
      "type": "Property",
      "value": "aaa"
    }
}
```

thomasBousselin:
https://github.com/stellio-hub/stellio-context-broker/issues/1689


bobeal (approved):

bobeal:
> #1689

Way better to link the issue in the right panel (it will even automatically close the issue when the PR is merged)

bobeal:
and normalized the title of the PR

PR #1773, opened 2026-06-15
fix: simplified temporal representation of a deleted languageMap

thomasBousselin:
I am not sure why this existed in the first place. But it was the reason behind the issue in the simplified temporal representation of a deleted languageMap. (All languageMap related test of the test suite still pass)

Note: the value parameter of addDeletedAttributeInstance could be removed entirely.

### DIFF
--- a/search-service/src/main/kotlin/com/egm/stellio/search/entity/model/Attribute.kt
+++ b/search-service/src/main/kotlin/com/egm/stellio/search/entity/model/Attribute.kt
@@ -1,9 +1,7 @@
 package com.egm.stellio.search.entity.model
 
 import com.egm.stellio.shared.model.ExpandedTerm
-import com.egm.stellio.shared.model.JSONLD_LANGUAGE_KW
 import com.egm.stellio.shared.model.JSONLD_NONE_KW
-import com.egm.stellio.shared.model.JSONLD_VALUE_KW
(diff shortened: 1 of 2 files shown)


thomasBousselin:
Analysis :
The compacted payload of 

```json
  ...
  "https://uri.etsi.org/ngsi-ld/hasLanguageMap": "urn:ngsi-ld:null"
``` 
or 
```json
  // ...
  "https://uri.etsi.org/ngsi-ld/hasLanguageMap": [
    {
      "@value": "urn:ngsi-ld:null"
    }
  ]
```
is 
```json
  // ...
  "languageMap": {
    "@none": "urn:ngsi-ld:null"
  }
```

That is why the spec specify it as the null representation of a deleted languageProperty.
And that is also why storing "urn:ngsi-ld:null" as value works.

Previous value was compacted into :
```json

 // ...
   "languageMap": {"@none": "[{'@value':'urn:ngsi-ld:null','@language':'@none'}]"}}

```
because of a remaining serializeObject. This mean that just removing the [serialize object](https://github.com/stellio-hub/stellio-context-broker/pull/1773/changes#diff-0cb125ccab509831e6be51d344aa6fd4a21005c806901b8db16211b46703e23dL153) (Attribute.kt line 153) would also fix the issue

bobeal (approved):

PR #1783, opened 2026-06-19
feat: support type wildcard

### LINKED ISSUE #1772: Allow broader local queries for Query Entities via POST
As of today it is not possible to query multiple entities only by their `id` using a post query entity operation. For instance, this query return an error: 

POST  `/ngsi-ld/v1/entityOperations/query`

```json
{
  "type": "Query",
  "entities": [
    {
      "id": "urn:ngsi-ld:IrrigationArea:621fbb4e-dead-431c-babc-ed6097e6a398"
    }
  ]
}
```

But this payload works:
```json
{
  "type": "Query",
  "entities": [
    {
      "id": "urn:ngsi-ld:IrrigationArea:621fbb4e-dead-431c-babc-ed6097e6a398",
      "type": "IrrigationArea"
    }
  ]
}
```

thomasBousselin:


bobeal:
I think the implementation is not totally correct.

I understand that the wildcard type thing only applies to the `EntitySelector` (because the `type` field is required in this data type and it is needed a way to say "I don't want any type restriction"), so it only applies to Query Entities via POST and in Subscriptions.

For Query Entities and Purge Entities, you can already say `local=true` and not specify a `type` query parameter. So the wildcard type does not apply in these operations.

thomasBousselin:
Yes, you can already say local=true, but type=* is an historic functionality. And is specifically specified for Query Entities and Purge Entities. (6.4.3.2-1, 6.4.3.3-1)

bobeal:
> Yes, you can already say local=true, but type=* is an historic functionality. And is specifically specified for Query Entities and Purge Entities.

`type=*` is not so historic, it has been added in 1.8. But, indeed, it is also specified for Query Entities and Purge Entities, even though I don't really see the point...

thomasBousselin:
> type=* is not so historic, it has been added in 1.8. But, indeed, it is also specified for Query Entities and Purge Entities, even though I don't really see the point...

Oh really, It would have make more sense if it was an historic feature. 

(Maybe because most people don't understand distribution and won't understand what local=true mean.)

search-service/src/main/kotlin/com/egm/stellio/search/authorization/permission/service/PermissionService.kt
```diff
@@ -517,13 +517,13 @@ class PermissionService(
         }
 
         //  targetTypeFilter also return permission targeting entity with this type
-        val targetTypeFilter = if (!permissionFilters.targetTypeSelection.isNullOrEmpty())
+        val targetTypeFilter = buildTypeQuery(permissionFilters.targetTypeSelection, "target_types")?.let {
             """
```
bobeal: the `.let {}` is no longer necessary?

search-service/src/main/kotlin/com/egm/stellio/search/entity/util/EntitiesQueryUtils.kt
```diff
@@ -39,7 +42,15 @@ fun composeEntitiesQueryFromGet(
     contexts: List<String>
 ): Either<APIException, EntitiesQueryFromGet> = either {
     val ids = queryParams.getFirst(QueryParameter.ID.key)?.split(",").orEmpty().toListOfUri().toSet()
-    val typeSelection = expandTypeSelection(queryParams.getFirst(QueryParameter.TYPE.key), contexts)
+    val localParameter = queryParams.getFirst(QueryParameter.LOCAL.key)?.toBoolean()
+    val (typeSelection, local) = // type=* mean local=true
```
bobeal:
```suggestion
    val (typeSelection, local) = // type=* implies local=true
```

search-service/src/main/kotlin/com/egm/stellio/search/entity/util/EntitiesQueryUtils.kt
```diff
@@ -137,6 +147,13 @@ fun composeEntitiesQueryFromPost(
     queryParams: MultiValueMap<String, String>,
     contexts: List<String>
 ): Either<APIException, EntitiesQueryFromPost> = either {
+    val localParameter = queryParams.getFirst(QueryParameter.LOCAL.key)?.toBoolean()
```
bobeal, line 152: `local` is marked as not implemented in `queryEntitiesViaPost`

search-service/src/test/kotlin/com/egm/stellio/search/entity/service/EntityServiceQueriesTests.kt
```diff
@@ -405,6 +406,31 @@ class EntityServiceQueriesTests : WithTimescaleContainer, WithKafkaContainer() {
             .contains(entity01Uri, entity02Uri)
     }
 
+    @Test
+    fun `it should retrieve entities without type restriction for a wildcard entity selector`() = runTest {
```
bobeal: can you adapt to the soon to come new naming?

search-service/src/test/kotlin/com/egm/stellio/search/entity/util/EntitiesQueryUtilsTests.kt
```diff
@@ -103,6 +104,21 @@ class EntitiesQueryUtilsTests {
         assertEquals(30, entitiesQuery.paginationQuery.limit)
     }
 
+    @Test
+    fun `it should interpret wildcard type as local query without type restriction`() = runTest {
```
bobeal: can you adapt to the soon to come new naming?

search-service/src/test/kotlin/com/egm/stellio/search/entity/util/EntitiesQueryUtilsTests.kt
```diff
@@ -381,6 +397,46 @@ class EntitiesQueryUtilsTests {
         }
     }
 
+    @Test
+    fun `it should interpret wildcard type in Query entity selectors as local query without type restriction`() =
```
bobeal: can you adapt to the soon to come new naming?

bobeal (commented):
- Add tests for the type=* + local=false returns 400 case? (and one for each endpoint in the Test Suite)
- In subscription service, it is somewhat linked with the `localOnly` flag that is not yet implemented. For consistency, it could be added (and return a 501 if set to `false`)

search-service/src/main/kotlin/com/egm/stellio/search/authorization/permission/service/PermissionService.kt
thomasBousselin: Yes, because buildTypeQuery can now receive a null typeSelection.

search-service/src/main/kotlin/com/egm/stellio/search/entity/util/EntitiesQueryUtils.kt
thomasBousselin: Yes, because the distribution is not yet implemented in query by post.

search-service/src/main/kotlin/com/egm/stellio/search/authorization/permission/service/PermissionService.kt
bobeal: yes, it is no longer necessary?

search-service/src/main/kotlin/com/egm/stellio/search/entity/util/EntitiesQueryUtils.kt
bobeal: Thanks, but I know that :)

The point is that this line will never be reached.

Two options then:
- Open the `local` (for Query Entities via POST) and `localOnly` (for Subscriptions) query params / member but return a 501 if called with a `false` value (the cleanest one)
- Don't open but don't process them at all, assuming it is local by default since dist ops are not implemented for both

bobeal (dismissed):

bobeal (approved):

PR #1787, opened 2026-06-21
refactor: homogenize test naming to function- and operation-prefixed conventions

bobeal:
## What

  Renames the test suite to a consistent naming convention and documents it, replacing
  the inconsistent mix of subject-less `it should …` names with names that state the
  unit under test.

  - **Unit / service / util tests** → `<functionUnderTest> should <behavior> [when <condition>]`
    (e.g. `mergeTemporalEntities should merge instances when datasetIds match`).
    The prefix is the literal Kotlin function, so tests are greppable from the production symbol.
  - **Controller / integration tests** → `<operation> should <result>`
    (e.g. `query temporal entity should return 400 if timerel is present without time`).

  The convention is codified in `.claude/rules/testing.md`; the old forms (`it should …`,
  `shouldn't`, names without `should`) are now banned "migrate on sight".

  ## Why

  Test names had drifted into several incompatible styles across modules (subject-less
  `it should …`, prose-prefixed, and a few function-prefixed), plus typos (`shoud`,
  `callSuscriber`) and a few names with no `should` clause at all. The function-prefixed
  form makes it trivial to jump from a production symbol to its tests and reads consistently.

  ## Scope

  - ~890 test renames across **67 test files** in all three modules
    (`subscription-service`, `shared`, `search-service`).
  - Fixes incidental typos surfaced along the way (`shoud`, `callSuscriber`, `simplfied`, `entiy`).
  - Where a long function prefix pushed a declaration past the 120-char Detekt limit, the
    predicate was tightened while preserving meaning.
  - Updated **7 Detekt baseline entries** whose keys referenced the old test-function names
    (otherwise previously-suppressed complexity/length findings resurface).

  No production code changes — tests only (plus the rules doc and baselines).

  ## Verification

  - `./gradlew :shared:test` and `./gradlew :search-service:test` pass in full.
  - `subscription-service` compiles and all non-container tests pass.
  - `detekt` is green on all three modules.
  - Repo-wide check: zero `it should …` / `shouldn't` / no-`should` test names remain.

  🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/search-service/src/test/kotlin/com/egm/stellio/search/authorization/subject/web/AnonymousUserHandlerTests.kt
+++ b/search-service/src/test/kotlin/com/egm/stellio/search/authorization/subject/web/AnonymousUserHandlerTests.kt
@@ -53,7 +53,7 @@ class AnonymousUserHandlerTests {
 
     @Test
     @WithAnonymousUser
-    fun `it should not authorize an anonymous to call the API`() {
+    fun `GET should return 401 for an anonymous user`() {
         webClient
             .get()
             .uri("/ngsi-ld/v1/entities/urn:ngsi-ld:Sensor:0022CCC")
--- a/search-service/src/test/kotlin/com/egm/stellio/search/authorization/subject/web/AnonymousUserHandlerTests.kt
+++ b/search-service/src/test/kotlin/com/egm/stellio/search/authorization/subject/web/AnonymousUserHandlerTests.kt
@@ -53,7 +53,7 @@ class AnonymousUserHandlerTests {
 
     @Test
     @WithAnonymousUser
-    fun `GET should return 401 for an anonymous user`() {
+    fun `GET requests should return 401 for an anonymous user`() {
         webClient
             .get()
             .uri("/ngsi-ld/v1/entities/urn:ngsi-ld:Sensor:0022CCC")
--- a/shared/src/test/kotlin/com/egm/stellio/shared/web/ExceptionHandlerTests.kt
+++ b/shared/src/test/kotlin/com/egm/stellio/shared/web/ExceptionHandlerTests.kt
@@ -13,7 +13,7 @@ class ExceptionHandlerTests {
     ).build()
 
     @Test
-    fun `it should raise an error of type BadRequestData if the request payload is not a valid JSON fragment`() {
+    fun `POST should return a BadRequestData error when the payload is not a valid JSON fragment`() {
         val invalidJsonLdPayload =
             """
             {
@@ -35,7 +35,7 @@ class ExceptionHandlerTests {
     }
 
     @Test
-    fun `it should raise an error of type LdContextNotAvailable if the context is not resolvable`() {
+    fun `POST should return an LdContextNotAvailable error when the context is not resolvable`() {
         val invalidJsonPayload =
             """
             {
--- a/shared/src/test/kotlin/com/egm/stellio/shared/web/ExceptionHandlerTests.kt
+++ b/shared/src/test/kotlin/com/egm/stellio/shared/web/ExceptionHandlerTests.kt
@@ -13,7 +13,7 @@ class ExceptionHandlerTests {
     ).build()
 
     @Test
-    fun `POST should return a BadRequestData error when the payload is not a valid JSON fragment`() {
+    fun `POST requests should return a BadRequestData error when the payload is not a valid JSON fragment`() {
         val invalidJsonLdPayload =
             """
             {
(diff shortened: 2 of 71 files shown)


shared/src/test/kotlin/com/egm/stellio/shared/web/ExceptionHandlerTests.kt
```diff
@@ -13,7 +13,7 @@ class ExceptionHandlerTests {
     ).build()
 
     @Test
-    fun `it should raise an error of type BadRequestData if the request payload is not a valid JSON fragment`() {
+    fun `POST should return a BadRequestData error when the payload is not a valid JSON fragment`() {
```
thomasBousselin: maybe 
```suggestion
    fun `POST requests should return a BadRequestData error when the payload is not a valid JSON fragment`() {
```

shared/src/test/kotlin/com/egm/stellio/shared/web/ExceptionHandlerTests.kt
```diff
@@ -35,7 +35,7 @@ class ExceptionHandlerTests {
     }
 
     @Test
-    fun `it should raise an error of type LdContextNotAvailable if the context is not resolvable`() {
+    fun `POST should return an LdContextNotAvailable error when the context is not resolvable`() {
```
thomasBousselin:
```suggestion
    fun `POST requests should return an LdContextNotAvailable error when the context is not resolvable`() {
```

shared/src/test/kotlin/com/egm/stellio/shared/web/TenantWebFilterTests.kt
```diff
@@ -35,7 +35,7 @@ class TenantWebFilterTests {
     }
 
     @Test
-    fun `it should return a NonexistentTenant error if the tenant does not exist`() {
+    fun `GET should return a NonexistentTenant error when the tenant does not exist`() {
```
thomasBousselin:
```suggestion
    fun `GET requests should return a NonexistentTenant error when the tenant does not exist`() {
```

shared/src/test/kotlin/com/egm/stellio/shared/web/TenantWebFilterTests.kt
```diff
@@ -48,7 +48,7 @@ class TenantWebFilterTests {
     }
 
     @Test
-    fun `it should find a non-default tenant and add the NGSILD-Tenant header in the response`() {
+    fun `GET should add the NGSILD-Tenant header when the tenant is non-default`() {
```
thomasBousselin:
```suggestion
    fun `GET request should add the NGSILD-Tenant header when the tenant is non-default`() {
```

shared/src/test/kotlin/com/egm/stellio/shared/web/TenantWebFilterTests.kt
```diff
@@ -59,7 +59,7 @@ class TenantWebFilterTests {
     }
 
     @Test
-    fun `it should fallback to the default tenant if none was provided in the request`() {
+    fun `GET should fall back to the default tenant when none is provided`() {
```
thomasBousselin:
```suggestion
    fun `GET requests should fall back to the default tenant when none is provided`() {
```

search-service/src/test/kotlin/com/egm/stellio/search/authorization/subject/web/AnonymousUserHandlerTests.kt
```diff
@@ -53,7 +53,7 @@ class AnonymousUserHandlerTests {
 
     @Test
     @WithAnonymousUser
-    fun `it should not authorize an anonymous to call the API`() {
+    fun `GET should return 401 for an anonymous user`() {
```
thomasBousselin:
```suggestion
    fun `GET requests should return 401 for an anonymous user`() {
```

bobeal: don't bother reading all of them one by one!

I'll run it once again on the PR to double check what has been missed

thomasBousselin: Yes, i checked rapidly and find these GET and POST strange.

bobeal: OK, it is only for some specific handler tests where it was unable to associate a corresponding API operation. Gonna fix them manually.

PR #1789, opened 2026-06-21
fix: handle previously deleted attributes when upserting new ones

### LINKED ISSUE #1788: Incorrect event type when creating an attribute over a previously deleted one
When creating an attribute and a previous one existed but had been deleted, the event is considered as an update and not as a creation.

As it is flagged as an update, `EntityEventService` tries to find the previous payload of the attribute and fails doing so since it is actually a creation. These errors are visible in the logs with such messages (coming from the `getAttributePayload` function):

```
Unable to find attribute https://vocab.egm.io/myAttribute (urn:ngsi-ld:Dataset:Some) in entity urn:ngsi-ld:Distribution:ABCD
```

And it also later fails in subscription service since it receives an event not providing the expected data.

bobeal:


search-service/src/main/kotlin/com/egm/stellio/search/entity/service/EntityAttributeService.kt
```diff
@@ -215,6 +216,7 @@ class EntityAttributeService(
         attributeMetadata: AttributeMetadata,
         createdAt: ZonedDateTime,
         attributePayload: ExpandedAttributeInstance,
+        existedPreviously: Boolean
```
thomasBousselin, line 219: Could be `isReplace`? (if not current naming is fine)

bobeal: it is not specific to an attribute replace.

the place I spotted it in the logs is:
- delete attribute
- the following query: update attribute (re-creating the attribute)

(not a good use of the API, but it exists)

PR #1794, opened 2026-06-28
refactor: rewrite entity attribute write path to improve performance

### LINKED ISSUE #1839: Concurrent writes to the same entity return 204 but are dropped from entity_payload
## Summary

Two clients writing concurrently to the same entity both get `204 No Content`, but one write is silently discarded from `entity_payload`. It is still present in `temporal_entity_attribute` and `attribute_instance_audit`, so the entity API serves an entity that contradicts the broker's own attribute tables. We hit this with rooms holding their sensor membership as a multi-attribute, each publisher appending its own `datasetId` instance of a `members` relationship: when the publishers restart together, a room comes up missing a sensor although that publisher logged a `204`.

## Environment

- Stellio 2.35.0 (`stellio/stellio-search-service:2.35.0`) and `latest-dev` (image built 2026-07-27, `develop` at `58e096e`) - both affected
- Repository `docker-compose.yml` + `docker-compose-dependencies.yml`, unchanged apart from host ports; dedicated stack, fresh volume
- Postgres `stellio/stellio-timescale-postgis:16-2.26.4-3.6`, Kafka `confluentinc/cp-kafka:8.1.0`, auth disabled

## Steps to reproduce

<details>
<summary>Two writers, 50 appends of distinct <code>datasetId</code> instances, then read back (Node 18+, no dependencies, core context only)</summary>

```js
const LD = (process.argv[2] || 'http://localhost:8080') + '/ngsi-ld/v1';
const CORE = 'https://uri.etsi.org/ngsi-ld/v1/ngsi-ld-core-context-v1.8.jsonld';
const H = { 'content-type': 'application/json',
    link: `<${CORE}>; rel="http://www.w3.org/ns/json-ld#context"; type="application/ld+json"` };
const E = `urn:ngsi-ld:Room:${Date.now()}`, N = 50;
const sensor = (n) => `urn:ngsi-ld:Sensor:${n}`;

const add = async (n) => (await fetch(`${LD}/entities/${encodeURIComponent(E)}/attrs`, {
    method: 'POST', headers: H,
    body: JSON.stringify({ members: { type: 'Relationship', object: sensor(n), datasetId: sensor(n) } }),
})).status;

await fetch(`${LD}/entities`, { method: 'POST', headers: H, body: JSON.stringify({ id: E, type: 'Room' }) });

const all = [...Array(N).keys()], statuses = new Set();
await Promise.all([0, 1].map(async (p) => {
    for (const n of all.filter((n) => n % 2 === p)) statuses.add(await add(n));
}));

await new Promise((r) => setTimeout(r, 2500));      // rule out read-visibility lag
const e = await (await fetch(`${LD}/entities/${encodeURIComponent(E)}`, { headers: H })).json();
const stored = [].concat(e.members ?? []).map((i) => i.object);
console.log(`statuses  ${[...statuses].join(' ')}`);
console.log(`stored    ${stored.length} of ${N}`);
console.log(`missing   ${all.filter((n) => !stored.includes(sensor(n))).map(sensor).join(', ') || 'none'}`);
```

</details>

Expected: all 50 instances stored, as they are when the identical writes are issued serially. That control run stores all 50 every time. Actual: one instance missing, in 6 of 7 runs on 2.35.0 and in 4 of 4 runs on `latest-dev`.

```
statuses  204
stored    49 of 50
missing   urn:ngsi-ld:Sensor:49
```

Immediately afterwards, with no further writes to that entity:

```
temporal_entity_attribute rows           50
entity_payload members array length      49

Sensor:49 in temporal_entity_attribute    1
Sensor:49 in attribute_instance_audit     1
Sensor:49 in entity_payload               0
```

## Root cause

As far as we can tell: `handleSuccessOperationActions` reads every attribute of the entity, then `updateState` rebuilds the whole `entity_payload.payload` from that snapshot and writes it back unconditionally (`EntityService.kt` L551-552 and L563-588, same lines in 2.35.0 and in `develop`). At `READ COMMITTED`, two appends that insert different `temporal_entity_attribute` rows never block each other, so each one reads a snapshot without the other and the second `UPDATE entity_payload` overwrites the first. No serialization error is raised, both requests return `204`, and nothing is logged.

This matches the rest of what we saw. Writes to different entities are unaffected, and so are two writers updating the same attribute, since they contend on one `temporal_entity_attribute` row. Two writers updating two different existing attributes hit the same race but leave a stale value in the payload rather than a missing one, giving `alpha=25` and `beta=24` from the API while both are `25` in `temporal_entity_attribute`, in 5 runs out of 5. And any later write to the entity restores a lost instance, because it rebuilds the payload from the normalised tables again, so only the last write of a concurrent burst stays broken.

Draft PR #1794 mentions removing "a concurrent lost-update bug where two simultaneous attribute modifications would race to overwrite the full payload", which may be the same issue - noting it here since it is still open and `latest-dev` reproduces.

bobeal:
Replace the O(N) per-write pattern — getAllForEntity() SELECT + full entity
deserialization + full payload replace — with a single, self-referential
returns the resulting entity via RETURNING.

This removes the N-attribute read on every write and eliminates a
concurrent lost-update bug where two simultaneous attribute modifications
would race to overwrite the full payload.

Specific changes:
  - upsert() now returns created_at and modified_at via RETURNING, avoiding
    a second SELECT to obtain the actual DB timestamps
  - patchEntityPayload() batches every distinct attribute name touched by
    an operation into one UPDATE, and reads the resulting entity straight
    from that statement's RETURNING clause instead of re-deriving it with
    a separate in-memory merge
  - each EntityAttributeService function pre-embeds sysAttrs into
    newExpandedValue before building the result
  - updateState() and buildJsonLdEntity() removed

Also fixes several issues found in review:
  - missing NGSI-LD spec conformance points for scope handling (5.6.2.4,
    5.6.4.4, 5.6.19.4)

### DIFF
--- a/search-service/src/main/kotlin/com/egm/stellio/search/entity/service/EntityService.kt
+++ b/search-service/src/main/kotlin/com/egm/stellio/search/entity/service/EntityService.kt
@@ -1,20 +1,21 @@
 package com.egm.stellio.search.entity.service
 
 import arrow.core.Either
-import arrow.core.flatMap
 import arrow.core.getOrElse
 import arrow.core.left
 import arrow.core.raise.either
+import arrow.core.raise.ensure
 import arrow.core.right
 import com.egm.stellio.search.authorization.permission.model.Action
 import com.egm.stellio.search.authorization.permission.model.getSpecificAccessPolicy
 import com.egm.stellio.search.authorization.permission.service.AuthorizationService
-import com.egm.stellio.search.common.util.deserializeAsMap
 import com.egm.stellio.search.common.util.deserializeExpandedPayload
 import com.egm.stellio.search.common.util.execute
+import com.egm.stellio.search.common.util.executeExpected
 import com.egm.stellio.search.common.util.oneToResult
+import com.egm.stellio.search.common.util.toJson
+import com.egm.stellio.search.common.util.toList
 import com.egm.stellio.search.common.util.toZonedDateTime
-import com.egm.stellio.search.entity.model.Attribute
 import com.egm.stellio.search.entity.model.AttributeOperationResult
 import com.egm.stellio.search.entity.model.EntitiesQueryFromGet
 import com.egm.stellio.search.entity.model.Entity
@@ -27,7 +28,7 @@ import com.egm.stellio.search.entity.model.OperationType.MERGE_ENTITY
 import com.egm.stellio.search.entity.model.OperationType.UPDATE_ATTRIBUTES
 import com.egm.stellio.search.entity.model.SucceededAttributeOperationResult
 import com.egm.stellio.search.entity.model.UpdateResult
-import com.egm.stellio.search.entity.model.getSucceededAttributesOperations
+import com.egm.stellio.search.entity.model.getSucceededOperations
 import com.egm.stellio.search.entity.model.hasSuccessfulResult
 import com.egm.stellio.search.entity.util.prepareAttributes
 import com.egm.stellio.search.entity.util.rowToEntity
@@ -37,6 +38,8 @@ import com.egm.stellio.search.scope.ScopeService
 import com.egm.stellio.search.temporal.model.AttributeInstance.TemporalProperty
 import com.egm.stellio.shared.model.APIException
 import com.egm.stellio.shared.model.AlreadyExistsException
+import com.egm.stellio.shared.model.BadRequestDataException
+import com.egm.stellio.shared.model.EXPANDED_ENTITY_CORE_MEMBERS
 import com.egm.stellio.shared.model.EXPANDED_ENTITY_SPECIFIC_MEMBERS
 import com.egm.stellio.shared.model.ExpandedAttribute
 import com.egm.stellio.shared.model.ExpandedAttributeInstance
@@ -45,15 +48,21 @@ import com.egm.stellio.shared.model.ExpandedAttributes
 import com.egm.stellio.shared.model.ExpandedEntity
 import com.egm.stellio.shared.model.ExpandedTerm
 import com.egm.stellio.shared.model.JSONLD_ID_KW
+import com.egm.stellio.shared.model.JSONLD_NONE_KW
 import com.egm.stellio.shared.model.JSONLD_TYPE_KW
+import com.egm.stellio.shared.model.NGSILD_DATASET_ID_IRI
+import com.egm.stellio.shared.model.NGSILD_MODIFIED_AT_IRI
 import com.egm.stellio.shared.model.NGSILD_SCOPE_IRI
 import com.egm.stellio.shared.model.NgsiLdEntity
-import com.egm.stellio.shared.model.addSysAttrs
+import com.egm.stellio.shared.model.ResourceNotFoundException
 import com.egm.stellio.shared.model.flattenOnAttributeAndDatasetId
 import com.egm.stellio.shared.model.toAPIException
 import com.egm.stellio.shared.model.toNgsiLdAttributes
 import com.egm.stellio.shared.queryparameter.PaginationQuery
+import com.egm.stellio.shared.util.ErrorMessages.DbQuery.OPERATION_NO_RESULT_MESSAGE
 import com.egm.stellio.shared.util.ErrorMessages.Entity.entityAlreadyExistsMessage
+import com.egm.stellio.shared.util.ErrorMessages.Scope.SCOPE_NOT_A_TARGET_ATTRIBUTE_MESSAGE
+import com.egm.stellio.shared.util.JsonLdUtils.buildNonReifiedTemporalValue
 import com.egm.stellio.shared.util.JsonUtils.serializeObject
 import com.egm.stellio.shared.util.getSubFromSecurityContext
 import com.egm.stellio.shared.util.ngsiLdDateTime
@@ -131,7 +140,7 @@ class EntityService(
                 sub,
                 ExpandedEntity(emptyMap()),
                 expandedEntityWithMetadata,
-                attrsOperationResult.getSucceededAttributesOperations()
+                attrsOperationResult.getSucceededOperations()
             )
         }
     }.fold(
@@ -258,7 +267,7 @@ class EntityService(
         scopeService.replace(ngsiLdEntity, replacedAt).bind()
 
         val operationResult = deleteOperationResults.plus(createOrReplaceOperationResult)
-        operationResult.getSucceededAttributesOperations()
+        operationResult.getSucceededOperations()
             .forEach {
                 val sub = getSubFromSecurityContext()
                 if (it.operationStatus == OperationStatus.DELETED)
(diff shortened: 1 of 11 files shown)


thomasBousselin:
The changes are quite huge impact so i asked an LLM to check for regression (only for scopes for now) and it found two regressions :

- Updating a scope send the expanded scope inside entity event, causing a ClassCastException when consumed by subscriptions.

- You can add the same scope two times ([/A] + [/B, /B] becomes [/A, /B, /B]).

I will add this cases as tests in the testsuite.

bobeal:
> The changes are quite huge impact so i asked an LLM to check for regression (only for scopes for now) and it found two regressions :
> 
>     * Updating a scope send the expanded scope inside entity event, causing a ClassCastException when consumed by subscriptions.
> 
>     * You can add the same scope two times ([/A] + [/B, /B] becomes [/A, /B, /B]).
> 
> 
> I will add this cases as tests in the testsuite.

Adding this to my list. I have other in-progress modifications for the scopes (current implementation is not aligned with the spec).

thomasBousselin:
For non scope regression i found :
- the same error when adding a list of type with multiple instance of the same value.
- modifying an attribute doesn't change the modifiedAt of the entity.payload in the database

bobeal:
New, final version that fixes the concurrent writes lost update bug (and the regressions raised in the comments). As said in the desciption, it also fixes some NGSI-LD spec conformance points for scope handling.

Unfortunately, it has no impact on the global performance of the attributes updates operations (actually, `pg_stat_statements` shows the SQL queries are very fast in both cases, well below 1ms in our load tests scenarios). I will update on this in https://github.com/stellio-hub/stellio-context-broker/discussions/1830.

bobeal:
@thomasBousselin

thomasBousselin:
I think there was still a regression when updating/or removing an attribute the entity modifiedAt was not changed.

bobeal:
> I think there was still a regression when updating/or removing an attribute the entity modifiedAt was not changed.

It was actually only missing when permanently deleting an attibute, fixed now.

Also added a few tests to ensure `modifiedAt` is correctly updated in the entity payload when updating an attribute.

search-service/src/main/kotlin/com/egm/stellio/search/entity/service/EntityService.kt
```diff
@@ -344,41 +353,42 @@ class EntityService(
...
+                    (
+                        SELECT array_agg(new_type ORDER BY ordinality)
+                        FROM unnest(:new_types::text[]) WITH ORDINALITY AS u(new_type, ordinality)
+                        WHERE new_type <> ALL(ep.types)
+                    ),
+                    '{}'::text[]
+                ) AS to_add
+                FROM entity_payload ep
+                WHERE ep.entity_id = :entity_id
+            )
```
thomasBousselin: Is this `With` needed?

search-service/src/main/kotlin/com/egm/stellio/search/entity/service/EntityService.kt
```diff
@@ -859,17 +835,156 @@ class EntityService(
...
+            }
         }
-        updateState(
-            entityId,
-            ngsiLdDateTime(),
-            entityAttributeService.getAllForEntity(entityId)
-        ).bind()
     }
+
+    private suspend fun patchEntityPayload(
```
thomasBousselin: name is to broad

search-service/src/main/kotlin/com/egm/stellio/search/entity/service/EntityService.kt
```diff
@@ -859,17 +835,156 @@ class EntityService(
...
+        val attributesByNames = operationResults
+            .filter { it.attributeName !in EXPANDED_ENTITY_CORE_MEMBERS }
+            .groupBy { it.attributeName }
+
+        val updatedPayload = mergeAttributesIntoPayload(entityId, attributesByNames, modifiedAt, modifiedAtPatch).bind()
+        ExpandedEntity(updatedPayload.deserializeExpandedPayload())
+    }
+
+    // For every distinct attribute touched by the operation, keeps that attribute's untouched instances,
+    // drops the touched ones, and appends the new/updated ones, all in a single UPDATE.
```
thomasBousselin: I had to read this comment three times to understand it.

search-service/src/main/kotlin/com/egm/stellio/search/entity/service/EntityService.kt
```diff
@@ -859,17 +835,156 @@ class EntityService(
...
+        entityId: URI,
+        attributesByNames: Map<ExpandedTerm, List<SucceededAttributeOperationResult>>,
+        modifiedAt: ZonedDateTime,
+        modifiedAtPatch: Json
+    ): Either<APIException, Json> {
+        val batch = attributesByNames.map { (attrName, results) -> buildMergeBatchEntry(attrName, results) }
+        // computes the merged (kept + new) instance array for each attribute
+        // used twice below:
+        //   - once to find attribute names whose merged array is now empty (to remove from payload)
+        //   - once for the others (to set on payload)
```
thomasBousselin:
```suggestion
```

search-service/src/main/kotlin/com/egm/stellio/search/entity/service/EntityService.kt
thomasBousselin: Also the fact that the kotlin variable is named the same as the SQL one might be confusing, especially since they does not necessarily contains the same elements.

search-service/src/main/kotlin/com/egm/stellio/search/entity/service/EntityService.kt
thomasBousselin: In that case using a `With` might be a good idea since this is used two times.

search-service/src/main/kotlin/com/egm/stellio/search/entity/service/EntityService.kt
```diff
@@ -859,17 +835,156 @@ class EntityService(
...
+        )
+            .bind("entity_id", entityId)
+            .bind("modified_at", modifiedAt)
+            .bind("batch", Json.of(serializeObject(batch)))
+            .bind("modified_at_patch", modifiedAtPatch)
+            .oneToResult { toJson(it["payload"]) }
+    }
+
+    // One entry for a given attribute name: the datasetIds whose existing instances must not be kept
+    // from the live payload (deleted or replaced by this operation), and the new/updated instances to append.
```
thomasBousselin: I don't get it.

search-service/src/main/kotlin/com/egm/stellio/search/entity/service/EntityService.kt
```diff
@@ -859,17 +835,156 @@ class EntityService(
...
+    private fun buildMergeBatchEntry(
```
thomasBousselin: So this function calculate which instance should be deleted and which one should be updated?

search-service/src/main/kotlin/com/egm/stellio/search/entity/service/EntityService.kt
```diff
@@ -859,17 +835,156 @@ class EntityService(
...
+        val excludedDatasetIds = (deletedDatasetIds + upsertedResults.map { it.datasetId })
+            .map { it?.toString() ?: JSONLD_NONE_KW }
+        return mapOf(
+            "attrName" to attrName,
+            "excludedDatasetIds" to excludedDatasetIds,
+            "newInstances" to upsertedResults.map { it.newExpandedValue }
+        )
+    }
+
+    // Only used for permanent deletion of an attribute (just drop the attribute from the entity payload)
```
thomasBousselin:
```suggestion
    // Only used for permanent deletion of an attribute
```

thomasBousselin: Name is clear.

search-service/src/main/kotlin/com/egm/stellio/search/scope/ScopeService.kt
```diff
@@ -263,83 +249,121 @@ class ScopeService(
...
                     attributeName = NGSILD_SCOPE_IRI,
                     operationStatus = OperationStatus.CREATED,
-                    newExpandedValue = mapOf(NGSILD_SCOPE_IRI to scopes.toList())
+                    newExpandedValue = mapOf(NGSILD_SCOPE_IRI to scopes.map { mapOf(JSONLD_VALUE_KW to it) })
+                )
+            }.bind()
+    }
+
+    @Transactional
+    internal suspend fun performAppend(
```
thomasBousselin: So this complex query is to avoid duplicating the scope when there is two concurrent append? It may be easier to add a lock on the scope (and check that no lock exist) when we get the currentScopes? (same comment on the complex type update i guess)

PR #1811, opened 2026-07-12
feat: support NGSI-LD transient storage of entities and attributes

### LINKED ISSUE #1661: Transient Storage of Entities and Attributes
See https://cim.etsi.org/NGSI-LD/official/clause-4.html#4.22

bobeal:
Implements clause 4.22 of the NGSI-LD spec (v1.9): entities and attributes can now carry a client-settable `expiresAt` system property, and a periodic job garbage-collects data whose expiry has passed.

- Add `expiresAt` parsing/validation (must be a future date-time) to `NgsiLdEntity` and every `NgsiLdAttributeInstance` subtype, and register it as a core/sysattr member alongside createdAt/modifiedAt.
- Persist `expires_at` on `entity_payload` and `temporal_entity_attribute` (V0_55 migration), and thread it through entity/attribute creation, update and query paths.
- Add `TransientDataGarbageCollectionJob`, a `@Scheduled` per-tenant job (interval configurable via `search.transient-data.gc.interval`) that purges expired entities and attributes.
- CSR federation: propagate an entity-level expiresAt down to its attributes (4.5.5.2) and resolve conflicting expiresAt values across local/remote sources when merging entities, including temporal merges (4.5.5.3).


search-service/src/main/kotlin/com/egm/stellio/search/entity/job/TransientDataGarbageCollectionJob.kt
```diff
@@ -0,0 +1,61 @@
...
+ */
+@Component
+class TransientDataGarbageCollectionJob(
+    private val entityService: EntityService,
+    private val applicationProperties: ApplicationProperties
+) {
+
+    private val logger = LoggerFactory.getLogger(javaClass)
+
+    @Scheduled(fixedRateString = $$"${search.transient-data.gc.interval:3600000}")
```
thomasBousselin: If there is multiple search-service in parallel there will be two of this jobs running at the same time. (Not sure if this will cause issues since the function is idempotent)

I don't know what is the best way for handling this. A way I fixed this issue in the past was to have a spring profile "main" that will be the only one running scheduled job.

This issue is also in subscription-service for sendTimeIntervalNotification, but it makes less sense to have multiple subscription-service running in parallel. We will face the issue again with service execution (maybe it makes more sense to tackle the with service execution).

search-service/src/main/kotlin/com/egm/stellio/search/csr/util/ContextSourceUtils.kt
```diff
@@ -40,23 +43,78 @@ typealias AttributeByDatasetId = Map<String?, CompactedAttributeInstance>
...
+        mergedEntity: MutableMap<String, Any>,
+        sourceEntities: List<CompactedEntity>
+    ) {
+        val expiresAtValues = sourceEntities.map { it[NGSILD_EXPIRES_AT_TERM] as? String }
+        if (expiresAtValues.isEmpty()) return
+        if (expiresAtValues.any { it == null })
+            mergedEntity.remove(NGSILD_EXPIRES_AT_TERM)
+        else
+            mergedEntity[NGSILD_EXPIRES_AT_TERM] = expiresAtValues.filterNotNull()
+                .maxWith { a, b -> ZonedDateTime.parse(a).compareTo(ZonedDateTime.parse(b)) }
```
thomasBousselin, line 87: Can we do directly a maxWith (that treat the null case) to avoid going through the list 3 times.

search-service/src/main/kotlin/com/egm/stellio/search/csr/util/ContextSourceUtils.kt
```diff
@@ -40,23 +43,78 @@ typealias AttributeByDatasetId = Map<String?, CompactedAttributeInstance>
...
+                            ?: run { mergedEntityMap[id] = entity.toMutableMap() }
+                    }
+                    null
+                }.leftOrNull()
+            }.toNonEmptyListOrNull()
+
+        val sourceEntitiesById = (localEntities + remoteEntitiesWithExpiresAtPropagated.flatMap { it.first })
+            .groupBy { it[NGSILD_ID_TERM] }
+        mergedEntityMap.forEach { (id, mergedEntity) ->
+            resolveEntityExpiresAtConflict(mergedEntity, sourceEntitiesById[id] ?: emptyList())
```
thomasBousselin: I would see the getMergeNewValues function as a better place to add this logic. Would it work there?

search-service/src/main/kotlin/com/egm/stellio/search/csr/util/ContextSourceUtils.kt
```diff
@@ -236,6 +342,9 @@ object ContextSourceUtils {
                 currentValue == null -> value
                 key == NGSILD_ID_TERM || key == JSONLD_CONTEXT_KW -> currentValue
                 key == NGSILD_TYPE_TERM -> mergeTypeOrScope(currentValue, value)
+                // includes expiresAt: its final entity-level value is (re)computed from all sources by
+                // resolveEntityExpiresAtConflict once every CSR has been merged in (see 4.5.5.3), so whichever
+                // value the "earliest wins" rule below picks for it here is only a placeholder
```
thomasBousselin:
```suggestion
                // expiresAt will be overwritten by resolveEntityExpiresAtConflict (4.5.5.3)
```

search-service/src/main/kotlin/com/egm/stellio/search/csr/util/ContextSourceUtils.kt
```diff
@@ -88,7 +152,10 @@ object ContextSourceUtils {
             val currentValue = currentEntity[key]
             when {
                 currentValue == null -> value
-                key == NGSILD_ID_TERM || key == JSONLD_CONTEXT_KW -> currentValue
+                // expiresAt is a plain member here, not a merge target: its final entity-level value is
+                // (re)computed from all sources by resolveEntityExpiresAtConflict once every CSR has been merged
+                // in (see 4.5.5.3), so whichever value is kept in this intermediate step is only a placeholder
```
thomasBousselin:
```suggestion
                // expiresAt will be overwritten by resolveEntityExpiresAtConflict
```

search-service/src/main/kotlin/com/egm/stellio/search/entity/job/TransientDataGarbageCollectionJob.kt
```diff
@@ -0,0 +1,61 @@
...
+    fun purgeTransientData() {
+        applicationProperties.tenants.forEach { tenantConfiguration ->
+            runBlocking {
+                mono {
```
thomasBousselin, line 31: Is the goal to run the purge in parrallel

search-service/src/main/kotlin/com/egm/stellio/search/entity/job/TransientDataGarbageCollectionJob.kt
thomasBousselin: Could be a cron because we will probably want to execute it at some point during the night.

bobeal: Having multiple search services won't be an issue. First arriving will delete the expired entities, others will found nothing.

It is more a problem for subscription service since the same notification will be sent multiple times.

search-service/src/main/kotlin/com/egm/stellio/search/csr/util/ContextSourceUtils.kt
bobeal: - `maxWith` will not tell that there is at least one null element in the list (and `ZoneDateTime.parse` raises a NPE if an element is null)
- when in the `else`, we know that all elements are not null. but if I remove the `filterNotNull`, I have to `!!` the result. I tend to prefer keeping the `filterNotNull`

search-service/src/main/kotlin/com/egm/stellio/search/csr/util/ContextSourceUtils.kt
bobeal: Not out the box. `getMergeNewValues` and `getMergeTemporalNewValues` map values of the entities, they are not designed to change the structure of the entities (removing an `expiresAt` attribute).

As it quite of makes sense to keep this specific entity-level resolution conflict logic in a dedicated function, I'll keep it as is.

search-service/src/main/kotlin/com/egm/stellio/search/entity/job/TransientDataGarbageCollectionJob.kt
bobeal: I need the mono to write the tenant heading in its context and have everything works fine after that point (basically rebinding into the Spring stack).

search-service/src/main/kotlin/com/egm/stellio/search/entity/job/TransientDataGarbageCollectionJob.kt
bobeal: Ah? I was more seeing something like every hour. But I anyway changed it to a cron, it indeed gives more flexibility.

search-service/src/main/kotlin/com/egm/stellio/search/csr/util/ContextSourceUtils.kt
thomasBousselin: `getMergeNewValues` is designed to return the fusion of the two attributes.
I find it strange to use a second function that have to go back into each attribute to modify them, basically fixing what the first merge missed.

Especially since this function is already complex and adding this logic here doesn't help.

PR #1837, opened 2026-07-29
fix: performance issues on the detailed versions of discovery endpoints

### LINKED ISSUE #1836: fix: `?details=true` timeout on /types and /attributes endpoints due to duplicate row explosion
## Summary

`GET /ngsi-ld/v1/types?details=true` and `GET /ngsi-ld/v1/attributes?details=true` can time out (>60s) on datasets with a moderate number of entities, because the underlying queries in `EntityTypeService.getEntityTypes` and `AttributeService.getAttributeDetails` return one row per *(entity, attribute)* pair instead of one row per distinct *(type, attribute)* / *(attribute, type)* pair.

## Symptom

- SQL executes in a few seconds and returns millions of rows (one row per entity × attribute, not deduplicated).
- In-process handling of those rows (JSON-LD term compaction via `compactTerm`) then dominates request time, since each duplicate row is individually compacted before being discarded — `compactTerm` runs a full Titanium `JsonLd.compact` call per invocation.
- The request exceeds the gateway's timeout budget and returns an error to the client instead of a response.

## Root cause

1. **Missing `DISTINCT`**: both queries join `entity_payload` to `temporal_entity_attribute` and `unnest(types)`
   without `DISTINCT`, so row count scales with total entity count rather than distinct type/attribute cardinality.
2. **Compact-then-dedup**: term compaction happened on the full duplicate-laden result before deduplicating
   (`.toSet()` was applied *after* compaction), so every duplicate row paid for a full compaction call before being
   discarded.
3. **No cross-row memoization**: an attribute/type shared across many rows was compacted once per occurrence rather than once per distinct term in the whole response.

## Affected endpoints

- `GET /ngsi-ld/v1/types?details=true` (`EntityTypeService.getEntityTypes`)
- `GET /ngsi-ld/v1/attributes?details=true` (`AttributeService.getAttributeDetails`)

## Proposed fix

- Add `DISTINCT` to both queries so they return distinct pairs directly from the database (mirrors the existing
  `getEntityTypeList`/`getAttributeList` queries, which already use `DISTINCT` and don't exhibit this issue).
- Deduplicate before compacting, using a request-scoped cache so each distinct term is compacted at most once.

No API/behavioural change — response shape and content are unaffected, this is a pure performance fix.

bobeal:


shared/src/main/kotlin/com/egm/stellio/shared/util/JsonLdUtils.kt
```diff
@@ -379,6 +379,17 @@ object JsonLdUtils {
...
+    /**
+     * Compact a term using the provided context, memoizing the result in the caller-provided cache so a term
+     * repeated across a batch is only compacted once.
+     */
+    fun compactTermWithCache(
+        term: ExpandedTerm,
+        contexts: List<String>,
+        cache: MutableMap<ExpandedTerm, String>
+    ): String =
+        cache.getOrPut(term) { compactTerm(term, contexts) }
```
thomasBousselin, line 391: compactTerm is a fast function. Did you still had the performance issue with the distinct(types) and before adding this cache?

bobeal: "compactTerm is a fast function" 😲 definitely no...

thomasBousselin (approved):

PR #1838, opened 2026-07-29
feat: add timeout for all requests

thomasBousselin:
This allow configuration of a timeout in stellio. This is mainly to avoid cases when reverse proxy returns timeout but stellio kept running "indefinitely" in the background.

I wonder if we should define a different timeout for consumption and provision operation?


bobeal (commented):
- it will stop but will it stop any underlying SQL running forever (which was the main initial point)?
- no need to do a specific thing for subscription service (it is not really concerned anyway)

PR #1842, opened 2026-07-30
feat: add database transaction timeout after 50s for all requests

thomasBousselin:
Also added a behavior to @Transactionnal so it returns its error in the Either (when present) instead of Throwing.

Useful here for timeout in this PR but also for batch operation to allow the handling of database errors into a 207.


search-service/src/test/kotlin/com/egm/stellio/search/entity/service/EntityServiceTests.kt
```diff
@@ -187,7 +187,11 @@ class EntityServiceTests : WithTimescaleContainer, WithKafkaContainer() {
     }
 
     @Test
-    fun `createEntity should not create an entity if it contains an NGSI-LD Null value`() = runTest {
+    fun `createEntity should returns the error found in createAttributes`() = runTest {
```
thomasBousselin: This test was broken.
The request failed with a mockException that was catch by toApiException() and changed into BadRequestData which made the test succeed.

docs/admin/misc_configuration_tips.md
```diff
@@ -34,6 +34,21 @@ If running Stellio from `docker-compose`, it can be configured in the environmen
       - SPRING_HTTP_CODECS_MAX-IN-MEMORY-SIZE=10485760
 ```
 
+## Configure the transaction timeout
+
+You can configure the default timeout for Spring-managed database transactions in `search-service` and
+`subscription-service` by setting the `application.transaction-timeout` property. The default timeout is 40 seconds.
```
bobeal: since the default timeout for HTTP requests is 1m, I would expect something closer to this threshold for the default DB timeout.

search-service/src/test/kotlin/com/egm/stellio/search/entity/service/EntityServiceTests.kt
bobeal:
```suggestion
    fun `createEntity should return errors found when creating attributes`() = runTest {
```

shared/src/main/kotlin/com/egm/stellio/shared/config/TransactionalEitherConfiguration.kt
```diff
@@ -0,0 +1,138 @@
...
+    private fun Throwable.isTimeout(): Boolean =
+        this is QueryTimeoutException ||
+            this is TransactionTimedOutException ||
+            this is R2dbcTimeoutException ||
+            this is R2dbcException && sqlState == QUERY_CANCELED_SQL_STATE
+
+    companion object {
+        private val logger = LoggerFactory.getLogger(TransactionalEitherInterceptor::class.java)
+
+        private const val QUERY_CANCELED_SQL_STATE = "57014"
```
bobeal: good to add a reference (link) to this SQL state definition

bobeal:
```suggestion
        private const val QUERY_CANCELLED_SQL_STATE = "57014"
```

shared/src/test/kotlin/com/egm/stellio/shared/config/DefaultTimeoutR2dbcTransactionManagerTests.kt
```diff
@@ -0,0 +1,78 @@
...
+import io.r2dbc.spi.ConnectionFactory
+import io.r2dbc.spi.Result
+import io.r2dbc.spi.Statement
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+import org.springframework.transaction.support.DefaultTransactionDefinition
+import reactor.core.publisher.Mono
+import java.time.Duration
+
+class DefaultTimeoutR2dbcTransactionManagerTests {
```
bobeal, line 16: tests names in this file do not follow the conventions.

shared/src/test/kotlin/com/egm/stellio/shared/config/TransactionalEitherConfigurationTests.kt
```diff
@@ -0,0 +1,137 @@
...
+import org.springframework.transaction.ReactiveTransaction
+import org.springframework.transaction.ReactiveTransactionManager
+import org.springframework.transaction.TransactionDefinition
+import org.springframework.transaction.TransactionSystemException
+import org.springframework.transaction.annotation.EnableTransactionManagement
+import org.springframework.transaction.annotation.Transactional
+import reactor.core.publisher.Mono
+
+@SpringJUnitConfig(TransactionalEitherConfigurationTests.TestConfiguration::class)
+class TransactionalEitherConfigurationTests @Autowired constructor(
```
bobeal, line 31: any way to also test an explicit rollback?

bobeal (commented):
I was also wondering if there was a not-too-complex way to test with a real DB running?

docs/admin/misc_configuration_tips.md
thomasBousselin: I think we want this timeout to trigger before the http request one. That way we don't do calculations for a request that has already been killed. Plus, this timeout would return a good ProblemDetails format where the http timeout will not.
And this concerns only the transaction, so we have to consider the time of expansion/compaction that would happen outside of the transaction.

shared/src/test/kotlin/com/egm/stellio/shared/config/TransactionalEitherConfigurationTests.kt
thomasBousselin: What do you mean?

thomasBousselin: i added search-service/src/test/kotlin/com/egm/stellio/search/common/tenant/DatabaseTransactionTimeoutTests.kt is it what you were looking for?

docs/admin/misc_configuration_tips.md
bobeal: But ... if there is a DB timeout, we return a response and there is nothing to compact?

thomasBousselin: Yes, but, if the timeout doesn't trigger and the transaction ends after 47 sec. You would have the compaction, and it will probably trigger the request timeout.

shared/build.gradle.kts
```diff
@@ -16,6 +16,9 @@ the<DependencyManagementExtension>().apply {
 }
 
 dependencies {
+    implementation("org.springframework:spring-aop")
+    implementation("org.springframework:spring-r2dbc")
```
bobeal, line 20: side note: I remember a time where adding this dependency in the shared module failed all the tests because it was searching for a working DB configuration. Strange that it works now.

as a consequence, probably some other configurations duplicated in search and subscription services could be moved to shared (in another issue)?

shared/src/test/kotlin/com/egm/stellio/shared/config/TransactionalEitherConfigurationTests.kt
bobeal: Part of what I was looking for :)

Partly related to the PR (since it modifies error handling), I was thinking about a test that would trigger a rollback of the current transaction (but not because of a timeout).

docs/admin/misc_configuration_tips.md
bobeal: 13 seconds of compaction? ok, it is expensive but not THAT expensive ;)

thomasBousselin: If the transaction took 40sec it means there will probably be a lot of data to process. If you are confident on the treatment time, I can bump the timeout to 50 or 55sec.

bobeal: let's have a deal at 50s

bobeal:
no reaction to the other open comments?

shared/src/test/kotlin/com/egm/stellio/shared/config/TransactionalEitherConfigurationTests.kt
thomasBousselin: Done.

shared/build.gradle.kts
thomasBousselin: Good idea. i will make a PR.

bobeal: Start by doing an issue, the PR is not urgent at all.

bobeal (approved):
Please don't merge before I release the 2.35.2 fix version (tomorrow).

PR #1846, opened 2026-08-03
fix: preserve the altitude of GeoProperty values (#1840)

### LINKED ISSUE #1840: Support the altitude (third coordinate) of GeoProperty values
## Summary

A GeoProperty whose GeoJSON geometry carries three coordinates is stored with only two. The altitude is dropped, the request returns `201`, and nothing reports the loss.

NGSI-LD does not require 3D support, so this is a request rather than a conformance problem. RFC 7946 section 3.1.1 does allow a position to be `[longitude, latitude, altitude]`, and Stellio accepts such a value today rather than rejecting it, so the coordinate is quietly discarded instead of either being kept or refused.

## Environment

- Stellio 2.35.0
- Postgres `stellio/stellio-timescale-postgis:16-2.26.4-3.6`, Kafka `confluentinc/cp-kafka:8.1.0`

## Steps to reproduce

```bash
BROKER=http://localhost:8080
CORE='https://uri.etsi.org/ngsi-ld/v1/ngsi-ld-core-context-v1.8.jsonld'
LINK="<$CORE>; rel=\"http://www.w3.org/ns/json-ld#context\"; type=\"application/ld+json\""

curl -s -X POST "$BROKER/ngsi-ld/v1/entities" \
  -H "content-type: application/json" -H "Link: $LINK" \
  -d '{
        "id": "urn:ngsi-ld:AltProbe:1",
        "type": "AltProbe",
        "location": { "type": "GeoProperty",
                      "value": { "type": "Point", "coordinates": [21.7, 38.2, 110.5] } }
      }'

curl -s "$BROKER/ngsi-ld/v1/entities/urn:ngsi-ld:AltProbe:1" -H "Link: $LINK"
```

Expected: `coordinates` reads back as `[21.7, 38.2, 110.5]`. Actual: the write returns `201` with no warning, and the entity reads back as

```json
"location": { "type": "GeoProperty",
              "value": { "type": "Point", "coordinates": [21.7, 38.2] } }
```

The value is already two-dimensional in `entity_payload`:

```json
"https://uri.etsi.org/ngsi-ld/hasValue": [ { "@value": "POINT (21.7 38.2)" } ]
```

## Where the coordinate is lost

The conversion to WKT in `GeoUtils.geoJsonToWkt` writes with a plain `WKTWriter()`, whose JTS default output dimension is 2, so the Z ordinate parsed by `GeoJsonReader` is discarded before storage. Keeping it would give `POINT Z (21.7 38.2 110.5)`. The matching read path, `GeoUtils.wktToGeoJson`, would need to emit the third coordinate as well.

Storage does not look like the constraint. The `geo_value` column is the dimension-agnostic PostGIS `geometry` type, and 3D values round-trip through it:

```sql
select ST_AsEWKT(ST_GeomFromEWKT('POINTZ(21.7 38.2 110.5)')) as ewkt,
       ST_NDims(ST_GeomFromEWKT('POINTZ(21.7 38.2 110.5)')) as ndims;

-- ewkt   POINT(21.7 38.2 110.5)
-- ndims  3
```

## Why it would be useful

Anything tracked with a height: drones and UAV corridors, cranes, building floors, indoor positioning. Because the write is accepted and acknowledged, a client cannot detect the loss without reading the entity back and comparing.

Storing the geometry as a `JsonProperty` does preserve all three coordinates and works as a stopgap, but it gives up GeoProperty semantics and geospatial queries:

```json
"location": { "type": "JsonProperty",
              "json": { "type": "Point", "coordinates": [21.7, 38.2, 110.5] } }
```

If 3D is out of scope, rejecting a three-coordinate position with a `BadRequestData` error would at least make the limitation visible to clients.

itsampras:
## Proposed changes

References #1840.

A GeoProperty whose GeoJSON geometry carries a third coordinate
(altitude, per RFC 7946 3.1.1) was accepted and acknowledged, but
stored with only two coordinates. `GeoUtils.geoJsonToWkt` converts
every geometry through a `WKTWriter` left at its default output
dimension of 2, which silently drops the Z ordinate before storage.

The fix asks for a 3D output dimension instead. `WKTWriter` only
emits the Z ordinate for geometries that actually carry one, so
existing 2D data and 2D writes are unaffected.

Nothing else needed to change:
- `GeoJsonReader` already parses the third ordinate.
- `GeoJsonWriter` (used by `wktToGeoJson`) already renders it back.
- The `geo_value` / `entity_payload` columns are the dimension
  agnostic PostGIS `geometry` type.
- Geoqueries are unaffected: PostGIS's 2D predicates
  (`ST_Distance`, `ST_Within`, `ST_Contains`, `ST_Intersects`,
  `ST_Equals`, `ST_Disjoint`, `ST_Overlaps`) ignore the Z ordinate
  rather than failing on it, confirmed by the added geoquery test
  case.

## Types of changes

-   [x] Bugfix (non-breaking change which fixes an issue)
-   [ ] New feature (non-breaking change which adds functionality)
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

-   [x] I have read the CONTRIBUTING doc
-   [x] I have signed the CLA
-   [x] I have added tests that prove my fix is effective or that my feature works
-   [x] I have added necessary documentation (if appropriate)
-   [x] Any dependent changes have been merged and published in downstream modules

## Further comments

`GeoUtilsTests` is new: 3D round trip for Point and LineString, both
conversion directions, and two regressions asserting 2D Point and
Polygon output is unchanged.

For the geoquery concern raised on the issue, `equals, Point, [100.0,
0.0, 50.0], 1` was added to the existing parameterized geoquery test.
It queries with a 3D point against the existing 2D `BeeHive:01`
location and asserts the match count is unchanged from the 2D-only
row already there.

Verified locally: `shared:test` (391/391), `shared:detekt` and
`search-service:detekt` (0 issues on both), and the full
`EntityServiceQueriesTests` class (153/153, including the new 3D
case).

### DIFF
--- a/docs/user/information_model.md
+++ b/docs/user/information_model.md
@@ -41,7 +41,7 @@ There are several property types:
 
 - `Property`: a string, number, date, time, datetime, or boolean
 - `JsonProperty`: valid JSON data
-- `GeoProperty`: a GeoJSON geometry (use `location` to represent the entity’s location). GeoJSON coordinates may include a third altitude ordinate, which Stellio preserves.
+- `GeoProperty`: a GeoJSON geometry (use `location` to represent the entity’s location). GeoJSON coordinates may include a third altitude ordinate, which Stellio preserves. Altitude is not taken into account by geoqueries.
 - `LanguageProperty`: text in multiple languages
 
 Example with different attributes:
--- a/docs/user/information_model.md
+++ b/docs/user/information_model.md
@@ -41,7 +41,7 @@ There are several property types:
 
 - `Property`: a string, number, date, time, datetime, or boolean
 - `JsonProperty`: valid JSON data
-- `GeoProperty`: a GeoJSON geometry (use `location` to represent the entity’s location)
+- `GeoProperty`: a GeoJSON geometry (use `location` to represent the entity’s location). GeoJSON coordinates may include a third altitude ordinate, which Stellio preserves.
 - `LanguageProperty`: text in multiple languages
 
 Example with different attributes:
--- a/search-service/src/test/kotlin/com/egm/stellio/search/entity/service/EntityServiceQueriesTests.kt
+++ b/search-service/src/test/kotlin/com/egm/stellio/search/entity/service/EntityServiceQueriesTests.kt
@@ -356,6 +356,7 @@ class EntityServiceQueriesTests : WithTimescaleContainer, WithKafkaContainer() {
         "equals, Point, '[100.0, 0.0]', 1",
         "equals, Point, '[101.0, 0.0]', 0",
         "equals, Point, '[100.0, 0.0, 50.0]', 1",
+        "within, Polygon, '[[[95.0, -5.0, 50.0], [105.0, -5.0, 50.0], [100.0, 5.0, 50.0], [95.0, -5.0, 50.0]]]', 1",
         "intersects, Polygon, '[[[100.0, 0.0], [101.0, 0.0], [101.0, 1.0], [100.0, 1.0], [100.0, 0.0]]]', 1",
         "intersects, Polygon, '[[[101.0, 0.0], [102.0, 0.0], [102.0, -1.0], [101.0, 0.0]]]', 0",
         "disjoint, Point, '[101.0, 0.0]', 2",
--- a/search-service/src/test/kotlin/com/egm/stellio/search/entity/service/EntityServiceQueriesTests.kt
+++ b/search-service/src/test/kotlin/com/egm/stellio/search/entity/service/EntityServiceQueriesTests.kt
@@ -355,6 +355,7 @@ class EntityServiceQueriesTests : WithTimescaleContainer, WithKafkaContainer() {
         "contains, Polygon, '[[[90.0, 0.0], [100.0, 10.0], [110.0, 0.0], [100.0, -10.0], [90.0, 0.0]]]', 0",
         "equals, Point, '[100.0, 0.0]', 1",
         "equals, Point, '[101.0, 0.0]', 0",
+        "equals, Point, '[100.0, 0.0, 50.0]', 1",
         "intersects, Polygon, '[[[100.0, 0.0], [101.0, 0.0], [101.0, 1.0], [100.0, 1.0], [100.0, 0.0]]]', 1",
         "intersects, Polygon, '[[[101.0, 0.0], [102.0, 0.0], [102.0, -1.0], [101.0, 0.0]]]', 0",
         "disjoint, Point, '[101.0, 0.0]', 2",
--- a/shared/src/test/kotlin/com/egm/stellio/shared/util/GeoUtilsTests.kt
+++ b/shared/src/test/kotlin/com/egm/stellio/shared/util/GeoUtilsTests.kt
@@ -23,11 +23,22 @@ class GeoUtilsTests {
     }
 
     @Test
-    fun `geoJsonToWkt and wktToGeoJson should round-trip the altitude of a Point`() {
+    fun `geoJsonToWkt should produce WKT that wktToGeoJson can parse back`() {
         val geoJson = """{ "type": "Point", "coordinates": [21.7, 38.2, 110.5] }"""
 
         val wkt = geoJsonToWkt(geoJson).shouldSucceedAndResult()
 
         assertEquals(deserializeObject(geoJson), wktToGeoJson(wkt))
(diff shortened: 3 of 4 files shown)


itsampras:
I have read the CLA Document and I hereby sign the CLA

docs/user/information_model.md
```diff
@@ -41,7 +41,7 @@ There are several property types:
 
 - `Property`: a string, number, date, time, datetime, or boolean
 - `JsonProperty`: valid JSON data
-- `GeoProperty`: a GeoJSON geometry (use `location` to represent the entity’s location)
+- `GeoProperty`: a GeoJSON geometry (use `location` to represent the entity’s location). GeoJSON coordinates may include a third altitude ordinate, which Stellio preserves.
```
bobeal: can you also add a mention to say that the altitude is not taken into account by geoqueries?

search-service/src/test/kotlin/com/egm/stellio/search/entity/service/EntityServiceQueriesTests.kt
```diff
@@ -355,6 +355,7 @@ class EntityServiceQueriesTests : WithTimescaleContainer, WithKafkaContainer() {
         "contains, Polygon, '[[[90.0, 0.0], [100.0, 10.0], [110.0, 0.0], [100.0, -10.0], [90.0, 0.0]]]', 0",
         "equals, Point, '[100.0, 0.0]', 1",
         "equals, Point, '[101.0, 0.0]', 0",
+        "equals, Point, '[100.0, 0.0, 50.0]', 1",
```
bobeal, line 358: to be sure, can you also add a permutation with another geometry (the polygon below for instance)?

shared/src/test/kotlin/com/egm/stellio/shared/util/GeoUtilsTests.kt
```diff
@@ -0,0 +1,33 @@
...
+
+    @Test
+    fun `geoJsonToWkt should not add an altitude to a 2D Point`() {
+        val geoJson = """{ "type": "Point", "coordinates": [21.7, 38.2] }"""
+
+        assertEquals("POINT (21.7 38.2)", geoJsonToWkt(geoJson).shouldSucceedAndResult())
+    }
+
+    @Test
+    fun `geoJsonToWkt and wktToGeoJson should round-trip the altitude of a Point`() {
```
bobeal:
```suggestion
    fun `geoJsonToWkt should produce WKT that wktToGeoJson can parse back`() {
```

shared/src/test/kotlin/com/egm/stellio/shared/util/GeoUtilsTests.kt
```diff
@@ -0,0 +1,33 @@
...
+import com.egm.stellio.shared.util.JsonUtils.deserializeObject
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Test
+import org.springframework.test.context.ActiveProfiles
+
+@ActiveProfiles("test")
+class GeoUtilsTests {
+
+    @Test
+    fun `geoJsonToWkt should preserve the altitude of a Point`() {
```
bobeal, line 12: can you add a similar test for another geometry?

bobeal (changes requested):
Looks good. Only a few minor comments.

shared/src/test/kotlin/com/egm/stellio/shared/util/GeoUtilsTests.kt
Skorpinakos: yes, done in 93eb10052adaf50ca1e1fd5ccc271eef8af71059

search-service/src/test/kotlin/com/egm/stellio/search/entity/service/EntityServiceQueriesTests.kt
Skorpinakos: yes, although i had to use a different polygon since this one was already quite large and adding the Z parmeter in each point got me above to the MaxLineLength (120).

docs/user/information_model.md
Skorpinakos: yes, added in 79f3a157387d2dd3cf7c2e9b1314d1574197ce44

bobeal:
Merged, thanks for the contribution!

PR #1847, opened 2026-08-04
feat: Service registration CRUD

thomasBousselin:


shared/src/main/kotlin/com/egm/stellio/shared/queryparameter/GeoQuery.kt
```diff
@@ -76,6 +60,26 @@ data class GeoQuery(
             }
     }
 
+    private fun buildSqlFilter(targetGeometry: String): String {
```
bobeal: I don't get the changes in the geo functions, it needs to be explained.

thomasBousselin: The change was to apply a geoQ directly to column containing a geometry (location of the Registration)

But is no longer needed since we changed the location of the service registration to a geoQ. (you can't match two geoQ against each other)
So I have removed this change

bobeal: there was a change in the PostGIS functions used, which was my main worrying point.

thomasBousselin: The change in function was because when ST_Distance receive two geometry with SRID=4326 it returns an angle instead of meter. 
Probably why the current cast the location into the geography type and not into a geometry.

I think the ST_DistanceSphere is a bit better to avoid having two different types of geo data. But it is no longer part of this PR. I could make a new PR for location in ContextSourceRegistration since it will also need this change.

bobeal: yes, the cast to a `geography` type is on purpose.

anyway, not a priority and needs to be carefully evaluated beforehand.

shared/src/main/kotlin/com/egm/stellio/shared/util/ErrorMessages.kt
```diff
@@ -100,6 +100,18 @@ object ErrorMessages {
             "Context source $csrId returned an invalid payload, attribute is neither a List nor a Map: $attribute"
     }
 
+    object ServiceRegistrationErrorMessages {
```
bobeal: there is not this "ErrorMessages" suffix for other objects (and that would maybe fix the "too long line" detekt errors)

search-service/src/main/kotlin/com/egm/stellio/search/service/model/ServiceMode.kt
```diff
@@ -0,0 +1,16 @@
+package com.egm.stellio.search.service.model
+
+import com.fasterxml.jackson.annotation.JsonProperty
+
+enum class ServiceMode(val key: String) {
```
bobeal: can't live in ServiceRegistration.kt instead?

bobeal: normal that it is used in ServiceRegistration and ServiceInformation?

search-service/src/main/kotlin/com/egm/stellio/search/service/model/ServiceInformation.kt
```diff
@@ -0,0 +1,22 @@
...
+import com.egm.stellio.shared.model.BadRequestDataException
+import com.egm.stellio.shared.util.ErrorMessages.ServiceRegistrationErrorMessages.SERVICE_INFORMATION_NAME_REQUIRED_MESSAGE
+
+data class ServiceInformation(
+    val name: String = "",
+    val title: String? = null,
+    val description: String? = null,
+    val mode: ServiceMode? = null,
+    val input: InputInformation? = null,
+    val output: InputInformation? = null
```
bobeal, line 17: a bit chocked that output is of type input :)

search-service/src/main/kotlin/com/egm/stellio/search/service/service/ServiceRegistrationService.kt
```diff
@@ -0,0 +1,237 @@
...
+        filters: ServiceRegistrationFilters
+    ): Either<APIException, Int> {
+        val whereStatement = buildWhereStatement(filters)
+
+        return databaseClient.sql(
+            """
+            SELECT count(distinct registration.id)
+            FROM service_registration AS registration
+            LEFT JOIN jsonb_to_recordset(entities)
+                AS entity_info(id text, "idPattern" text, type text[]) ON true
```
bobeal, line 180: do you need this when you only want the count?

search-service/src/test/kotlin/com/egm/stellio/search/service/service/ServiceRegistrationServiceTests.kt
```diff
@@ -0,0 +1,219 @@
...
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.boot.test.context.SpringBootTest
+import org.springframework.data.r2dbc.core.R2dbcEntityTemplate
+import org.springframework.data.r2dbc.core.delete
+import org.springframework.test.context.ActiveProfiles
+import org.springframework.test.context.TestPropertySource
+
+@SpringBootTest
+@ActiveProfiles("test")
+@TestPropertySource(properties = ["application.authentication.enabled=false"])
```
bobeal: not sure you need this.

search-service/src/main/kotlin/com/egm/stellio/search/service/service/ServiceRegistrationService.kt
```diff
@@ -0,0 +1,237 @@
...
+            """.trimIndent()
+        )
+            .bind("limit", limit)
+            .bind("offset", offset)
+            .allToMappedList { rowToServiceRegistration(it) }
+    }
+
+    suspend fun getServiceRegistrationsCount(
+        filters: ServiceRegistrationFilters
+    ): Either<APIException, Int> {
```
bobeal, line 172: not new but wondering why getXXX returns a List but getXXXCount returns an Either 🤔

search-service/src/test/kotlin/com/egm/stellio/search/service/web/ServiceRegistrationHandlerTests.kt
```diff
@@ -0,0 +1,260 @@
...
+            .exchange()
+            .expectStatus().isNoContent
+
+        coVerify {
+            serviceRegistrationService.delete(registrationId)
+        }
+    }
+
+    @Test
+    fun `retrieve should propagate a not found error`() = runTest {
```
bobeal: move it next to other retrieve tests

search-service/src/main/kotlin/com/egm/stellio/search/service/model/ServiceInformation.kt
thomasBousselin: Yes i tried to find a better name for the type (ArgumentDescription maybe) but with no luck

search-service/src/main/kotlin/com/egm/stellio/search/service/service/ServiceRegistrationService.kt
thomasBousselin: Yes because it is used in the filter whereStatement

search-service/src/main/kotlin/com/egm/stellio/search/service/model/InputInformation.kt
```diff
@@ -0,0 +1,35 @@
...
+
+import com.fasterxml.jackson.annotation.JsonProperty
+import java.math.BigDecimal
+
+data class InputInformation(
+    val type: InputInformationType,
+    val required: Boolean = false,
+    val properties: Map<String, InputInformation>? = null,
+    val minimum: BigDecimal? = null,
+    val maximum: BigDecimal? = null
```
bobeal, line 11: this looks like a deviation from what is in the GR?
- `minimum` and `maximum` are inside the `properties`
- a bit surprised by the recursivity in the `properties`

search-service/src/main/kotlin/com/egm/stellio/search/service/model/ServiceInformation.kt
bobeal: in the current state of the GR, I understand the `output` as only a single thing with a type, not a data structure like the `input`.

search-service/src/main/kotlin/com/egm/stellio/search/service/model/ServiceMode.kt
bobeal: reopened for the first unanswered comment

search-service/src/main/kotlin/com/egm/stellio/search/service/service/ServiceRegistrationService.kt
bobeal: wondering alone?

search-service/src/main/kotlin/com/egm/stellio/search/service/model/InputInformation.kt
thomasBousselin: possible input : 
```json
12.2
```

```json
{
  "property": {
       "subproperty": {
         "subsubproperty": 12.2
       }
  }
}
```

search-service/src/main/kotlin/com/egm/stellio/search/service/model/ServiceInformation.kt
thomasBousselin: in the GR the input are just example. And it is easier to put just string as examples. 
There is no reason an output could not be a complex type.

search-service/src/main/kotlin/com/egm/stellio/search/service/service/ServiceRegistrationService.kt
thomasBousselin: Because we use the oneToResult return an either and allToMappedList don't.

bobeal: Yes, I can see that by myself.

bobeal: But that does not really answer the inconsistency

search-service/src/main/kotlin/com/egm/stellio/search/service/model/InputInformation.kt
bobeal: - my first point is not answered
- this recursivity is already needed at this stage of the PoC?

search-service/src/main/kotlin/com/egm/stellio/search/service/model/ServiceInformation.kt
bobeal: indeed, it can be a complex type. but probably not with the same model as the input.

search-service/src/main/kotlin/com/egm/stellio/search/service/model/InputInformation.kt
```diff
@@ -0,0 +1,35 @@
+package com.egm.stellio.search.service.model
+
+import com.fasterxml.jackson.annotation.JsonProperty
+import java.math.BigDecimal
+
+data class InputInformation(
```
bobeal, line 6: at some point later, I think it will deserve a refactoring to a sealed typed hierarchy. in its current state, it is quite fuzzy since `type` can be many different things and the members of `InputInformation` are only relevant for some of the types.

search-service/src/main/kotlin/com/egm/stellio/search/service/model/ServiceMode.kt
thomasBousselin: Yes, i put it in ServiceInformation.

search-service/src/main/kotlin/com/egm/stellio/search/service/model/ServiceInformation.kt
thomasBousselin: Why not? They both describe what a json body can contain. And even if we don't need as much for output as for input, defining a "simpler" type would make the whole thing more complicated because you now have two different types to understand.

search-service/src/main/kotlin/com/egm/stellio/search/service/model/InputInformation.kt
thomasBousselin: > minimum and maximum are inside the properties

No, `minimum` and `maximum` are used when the type is `integer` (in the examples, it is always inside properties, but you can say that your root element is an `integer`).

> this recursivity is already needed at this stage of the PoC?

Why lose time thinking on other possible implementations when the recursive way was the obvious way to do it.

thomasBousselin:
This PR is part of the service-execution branch and it will be merged when the service-execution is more advanced.

PR #1851, opened 2026-08-06
fix: handle list-like values when merging Attributes

### LINKED ISSUE #1834: PATCH on a JsonProperty containing an array returns HTTP 500
I am trying to update the blueprint attribute of the following entity:

`PATCH /ngsi-ld/v1/entities/urn:ngsi-ld:UseCaseConfig:SmartFarm`

The request body has this structure:

```
{
  "blueprint": {
    "type": "JsonProperty",
    "json": [
      {
        "id": "urn:ngsi-ld:Crop:Template",
        "type": "Template"
      },
      {
        "id": "urn:ngsi-ld:Campaign:Template",
        "type": "Template"
      }
    ]
  }
}
```
The content of the json array has been simplified in this example.

Stellio returns HTTP status 500:
```
{
  "instance": "/ngsi-ld/v1/entities/urn:ngsi-ld:UseCaseConfig:SmartFarm",
  "status": 500,
  "title": "java.lang.ClassCastException",
  "type": "https://uri.etsi.org/ngsi-ld/errors/InternalError"
}
```
The attribute is not updated.

Stellio version: 2.35.0

bobeal:


search-service/src/test/kotlin/com/egm/stellio/search/entity/util/PatchAttributeTests.kt
```diff
@@ -714,6 +714,198 @@ class PatchAttributeTests {
                         }
                     }
                     """.trimIndent()
+                ),
+                // a JsonProperty holding a single-element array of an object on both sides is deep-merged
```
thomasBousselin: You don't have a test with two multi-element array. And the comment makes it look like a specific behavior for single element array.

search-service/src/main/kotlin/com/egm/stellio/search/entity/util/AttributeUtils.kt
```diff
@@ -273,32 +323,21 @@ fun mergePatch(
...
-                }
+            attrName == NGSILD_JSONPROPERTY_JSON -> {
+                target[attrName] = mergeJsonProperty(
+                    source[attrName]!![0] as Map<String, Any>,
+                    attrValue[0] as Map<String, Any>
+                ) ?: attrValue
+            }
+            attrName == NGSILD_VOCABPROPERTY_VOCAB || attrName == NGSILD_PROPERTY_VALUE -> {
+                // a Property or VocabProperty holding an array of values cannot be safely merge patched,
+                // so copy the whole value from the update
```
thomasBousselin: This comment no longer apply here.

search-service/src/main/kotlin/com/egm/stellio/search/entity/util/AttributeUtils.kt
```diff
@@ -197,35 +197,85 @@ private fun mergeLanguageMap(
...
+ * one-element arrays as needed. Returns `null` when either side's content is not safely mergeable (arrays with
+ * zero or several elements, or scalars), signalling to the caller that the update should be copied wholesale.
+ */
+private fun mergeJsonProperty(
+    sourceInstance: Map<String, Any>,
+    updateInstance: Map<String, Any>
+): List<Map<String, Any>>? {
+    val sourceContent = extractMergeableJsonMap(sourceInstance[JSONLD_VALUE_KW])
+    val updateContent = extractMergeableJsonMap(updateInstance[JSONLD_VALUE_KW])
+    if (sourceContent == null || updateContent == null)
```
thomasBousselin, line 206: Maybe it should not return null if sourceContent is null but update content is not.

thomasBousselin: Ok, returning null mean you will take the new instance. I would prefer if this is treated here rather than in the function calling it.

search-service/src/main/kotlin/com/egm/stellio/search/entity/util/AttributeUtils.kt
```diff
@@ -197,35 +197,85 @@ private fun mergeLanguageMap(
...
+        return null
+
+    val mergedContent = JsonMerger().merge(
+        JsonUtils.serializeObject(sourceContent),
+        JsonUtils.serializeObject(updateContent)
+    ).deserializeAsMap()
+
+    val nullKeys = updateContent.filter { (_, v) -> v == NGSILD_NULL }.keys
```
thomasBousselin, line 214: no applyNgsiLdNullRemoval?

search-service/src/main/kotlin/com/egm/stellio/search/entity/util/AttributeUtils.kt
```diff
@@ -197,35 +197,85 @@ private fun mergeLanguageMap(
...
+    // restore the shape (bare object vs. one-element array) that the update's `@value` had before merging
+    val rewrappedContent = if (updateInstance[JSONLD_VALUE_KW] is List<*>) listOf(filteredContent) else filteredContent
+
+    return listOf(updateInstance + (JSONLD_VALUE_KW to rewrappedContent))
+}
+
+/**
+ * Extracts the JSON object to merge/null-filter out of a JsonProperty's raw `@value` content: the value itself
+ * if it is already a JSON object, or the single element if it is a one-element array holding a JSON object.
+ */
```
thomasBousselin: Useless comment the function is really small and clear and do not need explanation.

search-service/src/main/kotlin/com/egm/stellio/search/entity/util/AttributeUtils.kt
```diff
@@ -197,35 +197,85 @@ private fun mergeLanguageMap(
 }
 
 /**
- * Removes from the merged map any keys that have an NGSI-LD Null value in the update.
+ * Merges a JsonProperty instance's `@value` content with the corresponding update, unwrapping/rewrapping
+ * one-element arrays as needed. Returns `null` when either side's content is not safely mergeable (arrays with
+ * zero or several elements, or scalars), signalling to the caller that the update should be copied wholesale.
+ */
```
thomasBousselin: The title of the function should be enough here. The comment is only needed right now because the behavior with null is counterintuitive. Which can probably be avoided

search-service/src/main/kotlin/com/egm/stellio/search/entity/util/AttributeUtils.kt
```diff
@@ -197,35 +197,85 @@ private fun mergeLanguageMap(
...
+        is List<*> if value.size == 1 && value[0] is Map<*, *> -> value[0] as Map<String, Any>
+        else -> null
+    }
+
+/**
+ * Merges a Property or VocabProperty instance with the corresponding update. Returns `null` when the update
+ * holds an array of values (`updateInstances.size > 1`), signalling to the caller that it is not safely
+ * mergeable and the update should be copied wholesale.
+ */
+private fun mergePropertyOrVocabProperty(
```
thomasBousselin, line 232: The title of the function should be enough here. The comment is only needed right now because the behavior with null is counterintuitive. Which can probably be avoided

thomasBousselin (approved):

PR #1853, opened 2026-08-10
Feature: add missing logic in service registration

thomasBousselin:
The diff is broken because I moved the registration code into the service.registration package. 

Hopefully I did it in two separate commits so you can review the two commits separately to avoid the issue.

### DIFF
--- a/search-service/src/main/kotlin/com/egm/stellio/search/service/registration/model/InputInformation.kt
+++ b/search-service/src/main/kotlin/com/egm/stellio/search/service/registration/model/InputInformation.kt
@@ -1,5 +1,13 @@
 package com.egm.stellio.search.service.registration.model
 
+import arrow.core.Either
+import arrow.core.raise.either
+import arrow.core.raise.ensure
+import com.egm.stellio.shared.model.APIException
+import com.egm.stellio.shared.model.BadRequestDataException
+import com.egm.stellio.shared.util.ErrorMessages.ServiceExecutionErrorMessages.invalidInputElementKeyMessage
+import com.egm.stellio.shared.util.ErrorMessages.ServiceExecutionErrorMessages.invalidServiceExecutionInputMessage
+import com.egm.stellio.shared.util.ErrorMessages.ServiceExecutionErrorMessages.missingServiceExecutionInputMessage
 import com.fasterxml.jackson.annotation.JsonProperty
 import java.math.BigDecimal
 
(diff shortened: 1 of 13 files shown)


search-service/src/main/kotlin/com/egm/stellio/search/service/registration/model/ServiceRegistration.kt
```diff
@@ -33,6 +34,7 @@ data class ServiceRegistration(
     val id: URI = "urn:ngsi-ld:ServiceRegistration:${UUID.randomUUID()}".toUri(),
     val type: String = NGSILD_SERVICE_REGISTRATION_TERM,
     val endpoint: URI,
+    val endpointMethod: HttpMethod = HttpMethod.POST,
```
bobeal, line 37: a deviation from the GR?

search-service/src/main/kotlin/com/egm/stellio/search/service/registration/model/InputInformation.kt
```diff
@@ -0,0 +1,157 @@
...
+
+data class InputInformation(
+    val type: InputInformationType,
+    val required: Boolean = false,
+    val properties: Map<String, InputInformation>? = null,
+    val elements: Map<String, InputInformation>? = null,
+    val minimum: BigDecimal? = null,
+    val maximum: BigDecimal? = null,
+    val matchRegex: String? = null,
+    val maxSize: Int? = null
```
bobeal, line 22: I don't find these definitions anywhere in the GR

thomasBousselin: Yes, 
We discussed about those changed and decided that we needed to add at least a way to say :
- string should match a regex,
- list or string has a maximum size
- and a way to describe the elements of a list.

search-service/src/main/kotlin/com/egm/stellio/search/service/registration/model/ServiceRegistration.kt
thomasBousselin: Yes, there is no way to specify what http method to use in the GR.

bobeal (dismissed):

bobeal:
Don't know what you did here. I approved it but you force pushed some things and now, it does not even build.

thomasBousselin:
I think it is the rebase on the changes from `feature/service-registration`. It is not really important since it will be merge with the rest of the service-execution.

thomasBousselin:
This PR is part of the service-execution branch and it will be merged when the service-execution is more advanced.

PR #1855, opened 2026-08-12
feature: service execution

thomasBousselin:
The `service.execution.README.md` file contains the decision that differs (or were not clear) from the ServiceExecution group report.


search-service/src/main/kotlin/com/egm/stellio/search/service/execution/model/ServiceExecution.kt
```diff
@@ -0,0 +1,116 @@
...
+    val id: URI = "urn:ngsi-ld:ServiceExecution:${UUID.randomUUID()}".toUri(),
+    val type: String = NGSILD_SERVICE_EXECUTION_TERM,
+    val serviceId: URI,
+    val entityId: URI,
+    val entityType: String,
+    val input: Any,
+    /** Name-based service resolution is not implemented; [serviceId] is used instead. */
+    val serviceName: String? = null,
+    val executionStatus: ServiceExecutionStatus = ServiceExecutionStatus.PENDING,
+    val completion: Double? = null,
```
thomasBousselin: maybe not the best name. (progress?)

search-service/src/main/kotlin/com/egm/stellio/search/service/execution/model/ServiceExecution.kt
```diff
@@ -0,0 +1,116 @@
...
+    val output: Any? = null,
+    val responseStatusCode: Int? = null,
```
thomasBousselin, line 42: Wasn't sure where to put the statusCode of what the service returned, so it is here for now. I am definitely open to other idea.

shared/src/main/kotlin/com/egm/stellio/shared/util/ErrorMessages.kt
```diff
@@ -118,12 +118,27 @@ object ErrorMessages {
     }
 
     object ServiceExecutionErrorMessages {
```
bobeal: again this `ErrorMessages` suffix.

shared/src/main/kotlin/com/egm/stellio/shared/util/ErrorMessages.kt
```diff
@@ -118,12 +118,27 @@ object ErrorMessages {
...
+        fun serviceExecutionNotFoundMessage(id: URI) = "Service execution $id does not exist"
+        fun serviceExecutionAlreadyExistsMessage(id: URI) = "Service execution $id already exists"
+        fun serviceExecutionFailedToParseMessage(cause: String?) =
+            "Service execution cannot be parsed: $cause"
+        const val SERVICE_EXECUTION_CREATE_MEMBERS_MESSAGE =
```
bobeal:
```suggestion
        const val SERVICE_EXECUTION_RESERVED_MEMBERS_MESSAGE =
```

shared/src/main/kotlin/com/egm/stellio/shared/util/ErrorMessages.kt
```diff
@@ -118,12 +118,27 @@ object ErrorMessages {
...
+            "Members 'completion', 'output', 'responseStatusCode' and 'executionStatus' are reserved " +
+                "for reporting the service response"
+        const val SERVICE_EXECUTION_UPDATE_MEMBERS_MESSAGE =
+            "Service execution update is only for reporting 'completion', 'output', and 'executionStatus'"
+        const val SERVICE_EXECUTION_DELETE_OPTIONS_MESSAGE =
```
bobeal:
```suggestion
        const val SERVICE_EXECUTION_INVALID_OPTIONS_MESSAGE =
```

shared/src/main/kotlin/com/egm/stellio/shared/util/ErrorMessages.kt
```diff
@@ -118,12 +118,27 @@ object ErrorMessages {
...
 
     object ServiceExecutionErrorMessages {
+        fun serviceExecutionNotFoundMessage(id: URI) = "Service execution $id does not exist"
+        fun serviceExecutionAlreadyExistsMessage(id: URI) = "Service execution $id already exists"
+        fun serviceExecutionFailedToParseMessage(cause: String?) =
+            "Service execution cannot be parsed: $cause"
+        const val SERVICE_EXECUTION_CREATE_MEMBERS_MESSAGE =
+            "Members 'completion', 'output', 'responseStatusCode' and 'executionStatus' are reserved " +
+                "for reporting the service response"
+        const val SERVICE_EXECUTION_UPDATE_MEMBERS_MESSAGE =
```
bobeal:
```suggestion
        const val SERVICE_EXECUTION_INVALID_UPDATE_MESSAGE =
```

search-service/src/main/kotlin/com/egm/stellio/search/service/execution/model/ServiceExecution.kt
bobeal: yes, `progress` looks better.

search-service/src/main/kotlin/com/egm/stellio/search/service/execution/model/ServiceExecution.kt
bobeal: I guess it is mostly interesting in case of errors?

If so, what about keeping the whole response instead?

And not sure there is a better place than in the service execution object.

search-service/src/main/kotlin/com/egm/stellio/search/service/execution/service/ServiceExecutionService.kt
```diff
@@ -0,0 +1,160 @@
...
+
+    suspend fun delete(id: URI): Either<APIException, Unit> = either {
+        checkExistence(id).bind()
+        r2dbcEntityTemplate.delete<ServiceExecution>()
+            .matching(query(where("id").`is`(id)))
+            .execute()
+            .bind()
+    }
+
+    companion object {
```
bobeal: - this is not how other services are designed
- in the context of a Spring-managed service, for a function that is not a class-level utility, it does not really makes sense
- (I know you like `companion objects`, but don't overuse them)

search-service/src/main/kotlin/com/egm/stellio/search/service/execution/service/ServiceExecutionLauncher.kt
```diff
@@ -0,0 +1,99 @@
...
+                ServiceExecutionStatus.FAILURE
+            },
+            output = output,
+            responseStatusCode = responseStatusCode.value(),
+            modifiedAt = ngsiLdDateTime()
+        )
+    }
+
+    /** Invokes a service whose work is asynchronous and waits for its acknowledgement response. */
+    suspend fun invokeAsynchronousService(
```
bobeal, line 49:
```suggestion
    suspend fun invokeAsynchronous(
```

bobeal: or add the `Service` suffix on the other one

search-service/src/main/kotlin/com/egm/stellio/search/service/execution/service/ServiceExecutionLauncher.kt
```diff
@@ -0,0 +1,99 @@
...
+                serviceEndpointContactErrorMessage(serviceRegistration.id, serviceRegistration.endpoint),
+                exception.message
+            ).toProblemDetail()
+        }
+
+    private fun parseOutput(responseBody: String?): Any? =
+        responseBody
+            ?.takeUnless(String::isBlank)
+            ?.let { body ->
+                runCatching { DataTypes.mapper.readValue(body, Any::class.java) }
```
bobeal: cleaner to call `deserializeAs<Any>(body)`

search-service/src/main/kotlin/com/egm/stellio/search/service/execution/web/ServiceExecutionHandler.kt
```diff
@@ -0,0 +1,194 @@
...
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RequestParam
+import org.springframework.web.bind.annotation.RestController
+import reactor.core.publisher.Mono
+import java.net.URI
+
+@RestController
+@RequestMapping("/ngsi-ld/v1/services")
+@Validated
+class ServiceExecutionHandler(
```
bobeal, line 57: some endpoints contain business (validation) logic that should to the services.

search-service/src/main/kotlin/com/egm/stellio/search/service/execution/web/ServiceExecutionHandler.kt
```diff
@@ -0,0 +1,194 @@
...
+            serviceExecutionLauncher.cancelExecution(serviceExecutionId).bind()
+        if (DeleteOption.REMOVE in options)
+            serviceExecutionService.delete(serviceExecutionId).bind()
+
+        ResponseEntity.status(HttpStatus.NO_CONTENT).build<String>()
+    }.fold(
+        { it.toErrorResponse() },
+        { it }
+    )
+    companion object {
```
bobeal, line 172: weird.

search-service/src/main/kotlin/com/egm/stellio/search/service/execution/model/ServiceExecution.kt
thomasBousselin: yes it is mostly for the error case.
The error response is put in the output.

search-service/src/main/kotlin/com/egm/stellio/search/service/execution/web/ServiceExecutionHandler.kt
thomasBousselin: I moved `EXECUTION_RESULT_MEMBERS` into Service Execution, but kept `DeleteOption` here.

search-service/src/main/kotlin/com/egm/stellio/search/service/execution/web/ServiceExecutionHandler.kt
thomasBousselin: Okay, the difficulty here is that I need to validate the `ServiceExecution` before launching it with `ServiceExecutionLauncher`.

I see four possibilities:

1. Keep the current implementation, with the validation in the handler.
2. Have `ServiceExecutionService` perform the validation and call `ServiceExecutionLauncher`. (I always saw the responsabillity of the service to be mostly database oriented, so this seems a bit strange to me, but it is a fine solution.)
3. Have `ServiceExecutionLauncher` call `ServiceExecutionService`. (This would apply when you create the execution.)
4. Introduce a new component responsible for the validation and for calling both `ServiceExecutionService` and `ServiceExecutionLauncher` (An elegant solution, but it differs too much from the rest of the codebase.)

search-service/src/main/kotlin/com/egm/stellio/search/service/execution/service/ServiceExecutionLauncher.kt
thomasBousselin: added Service on the other one

search-service/src/main/kotlin/com/egm/stellio/search/service/execution/web/ServiceExecutionHandler.kt
thomasBousselin: After reflexion i would lean toward the third option.

bobeal: Yes, option 3 looks fine.

search-service/src/main/kotlin/com/egm/stellio/search/service/execution/model/ServiceExecution.kt
bobeal: I may be wrong but looking at the tests, it looks like `output` will contain the technical outputs (`"accepted": "true"` or the error responses), as well as the business results of the execution (thus sort of mixing the control plane and the data plane). I would dissociate them into two separate fields.

bobeal: it also implies that the control plane should be like a temporal property (recording when the current ouput has been received)

search-service/src/test/kotlin/com/egm/stellio/search/service/execution/service/ServiceExecutionLauncherTests.kt
```diff
@@ -0,0 +1,163 @@
...
+                .willReturn(okJson("\"done\""))
+        )
+
+        val successfulExecution = serviceExecutionLauncher.invokeService(execution, registration)
+
+        assertEquals("done", successfulExecution.output)
+        assertEquals(200, successfulExecution.responseStatusCode)
+        verify(
+            getRequestedFor(urlPathEqualTo("/invoke"))
+                .withRequestBody(equalTo("\"turn-on\""))
```
bobeal, line 93: A `GET` with a body is strongly unconventional. If we really need to support `GET` methods (do we?), query params are a better way to convey input (but input will have to be a simple flat object).

search-service/src/test/kotlin/com/egm/stellio/search/service/execution/web/ServiceExecutionHandlerTests.kt
```diff
@@ -0,0 +1,392 @@
...
+            .uri("$resourceUri/$executionId")
+            .exchange()
+            .expectStatus().isEqualTo(501)
+
+        coVerify { serviceExecutionLauncher.cancelExecution(executionId) }
+        coVerify(exactly = 0) { serviceExecutionService.delete(any()) }
+    }
+
+    @Test
+    fun `delete with remove and cancel options should cancel before removal`() = runTest {
```
bobeal: convoluted way to test what you want to test

search-service/src/main/kotlin/com/egm/stellio/search/service/execution/model/ServiceExecution.kt
thomasBousselin: You are right, it  contains the technical outputs until the asynchronous service call the patch to update the status and add a potential output himself. But it doesn't make sense to define this on our own.

I added this into the list of thing that we need to discuss before publishing the spec.

search-service/src/test/kotlin/com/egm/stellio/search/service/execution/service/ServiceExecutionLauncherTests.kt
thomasBousselin: > If we really need to support GET methods (do we?)

Yes. We can imagine a service whose job is just to ask for a value somewhere. 


I think we would need to define a way to send query params for any http method.  For now you can send them if you defined it in your registration.endpoint but it means you can't modify them for the execution.

There could be a syntax like this
```json
{
  "endpoint": "https://host/path/${id}?queryparam=${value}"
  }
```
Then you provide these arguments in the execution.

I add this to the list of open subject to discuss with everyone.

search-service/src/main/kotlin/com/egm/stellio/search/service/execution/web/ServiceExecutionHandler.kt
thomasBousselin: Done

PR #1802, opened 2026-07-04
chore: upgrade to PostgreSQL 18 and Timescale 2.29.2

bobeal:

PR #1861, opened 2026-08-20
feat(db): use a global DB timeout configuration

bobeal:


shared/src/main/kotlin/com/egm/stellio/shared/config/DefaultTimeoutR2dbcTransactionManager.kt
thomasBousselin, line 23: This behavior was define specifically to add a default timeout on transaction it probably doesn't make sense to keep it.

thomasBousselin, line 33: This function may be useless now

bobeal:
It is more messy than initially expected:
1. the new property initially introduced is called `transactionTimeout` but it adds a statement timeout, not a transaction timeout. it is misleading and should be renamed.
2. Spring TX exposes a `spring.transaction.default-timeout` configuration property. Could not this be used instead of coding our own thing? (we are actually more interested in transaction timeout than in statement timeout)
3. The timeout defined in the `@Transactional` annotation seems to be translated into a lock timeout (see R2dbcTransactionManager.java:257-259 → ExtendedTransactionDefinition:462-465), which is yet another thing.

so, @thomasBousselin, since you initiated this timeout move, I am reassigning this PR to you so you can dig into it.

thomasBousselin:
I am digging into it, the best way to do it might be to upgrade to postgres 17. where they [added `transaction_timeout`](https://www.postgresql.org/docs/17/release-17.html#RELEASE-17-SERVER-CONFIG) configuration in postgres directly.

bobeal:
> I am digging into it, the best way to do it might be to upgrade to postgres 17. where they [added `transaction_timeout`](https://www.postgresql.org/docs/17/release-17.html#RELEASE-17-SERVER-CONFIG) configuration in postgres directly.

Available: https://github.com/stellio-hub/stellio-context-broker/pull/1802

PR #1871, opened 2026-09-04
fix(subscription): remove unbounded fire-and-forget in entity event listener

### LINKED ISSUE #1866: subscription-service: EntityEventListenerService fires unbounded coroutines (no backpressure) → OOM under sustained entity-event load
## Summary

The subscription-service's Kafka listener (`EntityEventListenerService`) processes every `cim.entity._CatchAll` message by launching a **fire-and-forget coroutine on `Dispatchers.Default`** with no `await`, no bounded queue, and no backpressure. Under sustained entity-event load (our broker sees ~150 events/sec), the pending-coroutine queue + the full expanded-entity payload held in each dispatch grow without bound until the JVM hits `OutOfMemoryError: Java heap space`.

We confirmed this is the OOM mechanism by decompiling the shipped 2.36.0 jar (`com.egm.stellio.subscription.listener.EntityEventListenerService`), and verified the same code is still on `master`.

## Evidence

`subscription-service/src/main/kotlin/com/egm/stellio/subscription/listener/EntityEventListenerService.kt` (master):

```kotlin
@Component
class EntityEventListenerService(
    private val notificationService: NotificationService
) {
    private val logger = LoggerFactory.getLogger(javaClass)

    private val coroutineScope = CoroutineScope(Dispatchers.Default)

    @KafkaListener(topics = ["cim.entity._CatchAll"], groupId = "context_subscription")
    fun processMessage(content: String) {
        coroutineScope.launch {          // <-- fire-and-forget, returns immediately
            dispatchEntityEvent(content) //     acked before this completes
        }
    }
```

`dispatchEntityEvent` deserializes the entity event, then `handleEntityEvent` →
`notificationService.notifyMatchingSubscribers(...)` does R2DBC round-trips + an HTTP/MQTT
notification per matching subscription, all inside the launched coroutine.

Consequence: the Spring Kafka consumer acks each record as soon as `processMessage` returns
(instantly), so Kafka delivery is never throttled by downstream processing. Each record parks a
coroutine on `Dispatchers.Default`'s unbounded queue, each coroutine holds an `ExpandedEntity`
(full JSON-LD expansion, potentially large), and a burst or sustained rate fills the heap.

## Reproduction

1. Run `subscription-service` 2.36.0 (Spring Kafka, `@KafkaListener` as above) against a topic
   sustaining ~100+ entity events/sec.
2. Observe `jvm.memory.used` climb monotonically; under load the JVM throws
   `java.lang.OutOfMemoryError: Java heap space` from `Exception in thread "DefaultDispatcher-worker-*"`.
3. After OOM, the process does NOT exit (no `-XX:+ExitOnOutOfMemoryError`), so it GC-thrashes at
   ~400% CPU consuming its whole heap as a zombie. (We worked around this operationally by adding
   `-XX:+ExitOnOutOfMemoryError` + a container restart policy, but that only converts the OOM into
   a restart loop — it does not fix the root cause.)

Our live signature over 120s (production broker): ~18,000 events consumed across entity types,
top of which was `TrafficLight` at ~6,800 events/120s — a type with **no subscription** — i.e. the
listener pays the full fire-and-forget dispatch cost even when zero subscribers match.

## Suggested fix

Bound the concurrency instead of fire-and-forget. Options, in order of preference:

1. **Bounded dispatcher**: `private val coroutineScope = CoroutineScope(Dispatchers.Default.limitedParallelism(N))`
   (or a `Semaphore(N)`/bounded `Channel`) so a slow downstream applies backpressure and the queue
   cannot grow unbounded.
2. **Await the dispatch**: change `processMessage` to suspend and `blockUntilApplicable`/join the
   launched job before returning, so the Kafka consumer acks only after processing — restoring
   natural backpressure (at the cost of a slower consumer, which is generally desirable here).
3. Optionally also add a `spring.kafka.listener.concurrency`/`max.poll.records` tunable path, but
   note that `max.poll.records` alone does NOT bound the coroutine queue because `processMessage`
   returns immediately regardless.

### Impact
- Severity: high (service enters OOM / restart or GC-thrash loop under load).
- Our deployments: City OS production, Stellio 2.36.0.
- We have added `-XX:+ExitOnOutOfMemoryError` as a stopgap so the failure is a ~30s restart blip
  rather than an unbounded zombie, but the underlying defect remains.

bobeal:
`EntityEventListenerService` launched a coroutine per Kafka record with no bound and no await, so the consumer acked (and kept polling) before notification processing ever finished. Under sustained load this let pending work pile up unboundedly and drove the JVM to OOM (#1866).

`processMessage` now blocks the Kafka consumer thread until the event is fully processed, restoring backpressure. Tenant propagation moves from `Mono#contextWrite/subscribe` to `withContext(ReactorContext(...))`, the coroutine-native equivalent, so downstream tenant resolution keeps working without the extra detached async hop.

### DIFF
--- a/subscription-service/src/main/kotlin/com/egm/stellio/subscription/listener/EntityEventListenerService.kt
+++ b/subscription-service/src/main/kotlin/com/egm/stellio/subscription/listener/EntityEventListenerService.kt
@@ -1,8 +1,5 @@
 package com.egm.stellio.subscription.listener
 
-import arrow.core.Either
-import arrow.core.raise.either
-import com.egm.stellio.shared.model.APIException
 import com.egm.stellio.shared.model.AttributeCreateEvent
 import com.egm.stellio.shared.model.AttributeDeleteEvent
 import com.egm.stellio.shared.model.AttributeUpdateEvent
@@ -18,14 +15,13 @@ import com.egm.stellio.shared.util.JsonUtils.deserializeExpandedPayload
 import com.egm.stellio.shared.util.NGSILD_TENANT_HEADER
 import com.egm.stellio.subscription.model.NotificationTrigger
 import com.egm.stellio.subscription.service.NotificationService
-import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.launch
-import kotlinx.coroutines.reactor.mono
+import kotlinx.coroutines.reactor.ReactorContext
+import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.withContext
 import org.slf4j.LoggerFactory
 import org.springframework.kafka.annotation.KafkaListener
 import org.springframework.stereotype.Component
-import reactor.core.Disposable
+import reactor.util.context.Context
 import java.net.URI
 
 @Component
@@ -34,13 +30,11 @@ class EntityEventListenerService(
 ) {
     private val logger = LoggerFactory.getLogger(javaClass)
 
-    private val coroutineScope = CoroutineScope(Dispatchers.Default)
-
+    // block the Kafka consumer poll thread until the event is fully processed, so only ack
+    // (and poll for more records) once downstream processing has actually kept up (see issue #1866)
     @KafkaListener(topics = ["cim.entity._CatchAll"], groupId = "context_subscription")
-    fun processMessage(content: String) {
-        coroutineScope.launch {
-            dispatchEntityEvent(content)
-        }
+    fun processMessage(content: String) = runBlocking {
+        dispatchEntityEvent(content)
     }
 
     internal suspend fun dispatchEntityEvent(content: String) {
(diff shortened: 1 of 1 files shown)


thomasBousselin (approved):

limcheehow:
Hey @bobeal — thanks for the fast turnaround. We tested this on our production broker (Stellio 2.36.0, ~150 entity events/sec on `cim.entity._CatchAll`), cherry-picking the change onto 2.36.0 and running it under live load.

**It fixes the OOM.** Previously the heap climbed to ~10.8 GB and threw `OutOfMemoryError: Java heap space` every ~20–40 min, GC-thrashing the host in between. With the fix:
- heap flat at ~2.4 GiB across the soak, no second-order climb
- zero restarts, zero OOM events
- CPU steady instead of 300–1000% thrash

This resolves the recurring production OOM for us, so please ship it ASAP if you're comfortable.

One non-blocking observation: with the consumer now blocking, Kafka lag builds under sustained load (the firehose is dominated by event types with no matching subscriptions, e.g. TrafficLight churn). That's the "slower consumer" trade-off you flagged, not a defect in this change. A follow-up to partition `cim.entity._CatchAll` + `spring.kafka.listener.concurrency`, or short-circuit unsubscribed types before the expensive JSON-LD expansion, would recover the throughput — happy to open a separate issue for that if useful.

Thanks again.

bobeal:
Hi @limcheehow,

Good to hear it fixed the issue. Gonna merge this PR and release a new version (it was anyway planned for the next week).

PR #1875, opened 2026-09-14
fix: reject GeometryCollection geometries in GeoProperty values

### LINKED ISSUE #1850: Reject GeoJSON payloads containing a GeometryCollection
As per the NGSI-LD specification (4.6.3), `GeometryCollection` is not an authorized geometry.

However, Stellio does not proceed to such a check and thus it is possible to create a `GeoProperty` containing a `GeometryCollection`.

Add a check to forbid this and return a 400 Bad Request.

bobeal:


thomasBousselin:
Do you know what is the reason behind GeometryCollection rejection necessity? if it doesn't break anything i don't see the need to forbid it.

bobeal:
> Do you know what is the reason behind GeometryCollection rejection necessity? if it doesn't break anything i don't see the need to forbid it.

Don't know, it was in the specification before me :)

But I guess because it makes geoqueries a lot more complex and expensive.

thomasBousselin (approved):

PR #1879, opened 2026-09-17
Feature/list relationship and properties

thomasBousselin:


shared/src/main/kotlin/com/egm/stellio/shared/model/NgsiLdEntity.kt
```diff
@@ -607,10 +745,9 @@ fun checkAttributeHasNoForbiddenMembers(
     forbiddenMembers: Set<ExpandedTerm>
 ): Either<APIException, Unit> = either {
     forbiddenMembers.find {
```
thomasBousselin: Previous version was not returning the BadRequestException. (.bind() forgetted)

